PR #61, opened 2022-01-28
VBLOCKS-288: [Android] Ringtone is not going to the right device for inbound and outbound

mluotwilio:
## Submission Checklist

## Description

In Frontline app, inbound calls were always played on the earpiece, while outbound calls were always played on the speaker. The desired behaviors should be the opposite. This PR is the fix for that.

## Breakdown

- Replaced SoundPoolManager with MediaPlayerManager
- Rewrote the ringing code so that outbound calls are always played on the earpiece if no external devices are connected
- Changed the behavior of ringing of inbound calls such that speaker is used instead of earpiece
- Added comments to explain the changes

## Validation

- Manually tested the scenario where customer called worker (inbound) and observed that speaker was used for ringing instead of the earpiece
- Manually tested the scenario where worker called customer and observed that earpiece was used for the ringback tone instead of the speaker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ The Twilio Voice React Native SDK uses [Semantic Versioning](http://www.semver.o
   same event multiple times to the JS layer. This was most noticeable during
   development and testing but might rarely occur in normal usage.
 
+* Android ringtone issue [VBLOCKS-288]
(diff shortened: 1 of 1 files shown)


omerts:
@mluotwilio This seems to cause outgoing calls to only be in earpiece even when changing the audio device.

PR #131, opened 2023-02-14
[VBLOCKS-1434] AudioManager

mhuynh5757:
## Submission Checklist

## Description

Rework the AudioSwitchManager class singleton so it retains the audio device information.

## Breakdown

- Move audio device information into the resilient AudioSwitchManager singleton.
- Adjust JS layer to expect these changes.
- Add JS unit tests.

## Validation

- JS unit tests.
- Manually made calls with the test app.

## Additional Notes

N/A


android/src/main/java/com/twiliovoicereactnative/AudioManager.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * AudioManager singleton class. Maintains a persistent AudioSwitch object and listens for audio
+ * device changes. Generates UUIDs per audio device when the AudioSwitch library updates available
+ * audio devices.
+ */
+@RequiresApi(api = Build.VERSION_CODES.N)
+public class AudioManager {
```
kbagchiGWC, line 21: Can you choose a different name to not get confused with Android AudioManager? https://developer.android.com/reference/android/media/AudioManager

kbagchiGWC:
@mhuynh5757 Please fill out the PR description stating the issue, solution and how did you verify the changes including " I acknowledge that all my contributions will be made under the project's license." selection.

android/src/main/java/com/twiliovoicereactnative/AudioSwitchManager.java
```diff
@@ -1,28 +1,157 @@
...
+    return this;
+  }
+
+  /**
+   * Get the AudioSwitch. Note that the AudioSwitchManager is a singleton, and only instantiates an
+   * AudioSwitch once in the construction of the AudioSwitchManager. Therefore, there should only be one
+   * AudioSwitch object and it is effectively a singleton.
+   * @return The AudioSwitch singleton
+   */
+  public AudioSwitch getSwitch() {
```
kbagchiGWC, line 130: rename to `getAudioSwitch()`

kbagchiGWC (changes requested):
Left a minor comment. 
Have you tested reloading the app? Does the array of devices get populated correctly?

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 ## Fixes
 
 - Fixed an issue where some types on the `Call` and `Voice` classes were being incorrectly exported. Types and references to `addEventListener` are instead now correctly exported as `addListener`.
+- Fixed an issue where available audio devices were incorrectly emitted and returned by the SDK.
```
charliesantos, line 31: Were the audio devices incorrectly returned 100% of the time, or only on certain scenario? I think we should include that information.

api/voice-react-native-sdk.api.md
```diff
@@ -845,7 +845,7 @@ export interface Voice {
     // @internal (undocumented)
     emit(voiceEvent: Voice.Event, listener: (...args: any[]) => void): boolean;
     // @internal (undocumented)
-    emit(voiceEvent: Voice.Event.AudioDevicesUpdated, audioDevices: AudioDevice[], selectedDevice: AudioDevice): boolean;
+    emit(voiceEvent: Voice.Event.AudioDevicesUpdated, audioDevices: AudioDevice[], selectedDevice: AudioDevice | null): boolean;
```
charliesantos, line 848: Why not make `selectedDevice` optional instead?
>selectedDevice?: AudioDevice): boolean

charliesantos:
Left a couple of comments. I looked at the android related files as well but I'll leave the approval to our android expert :)

CHANGELOG.md
mhuynh5757: Only sometimes, good point. Should have only happened on Android when reloading bundles, probably only in development environments but maybe in prod environments.

api/voice-react-native-sdk.api.md
mhuynh5757: No strong reason for `null`, can be easily changed. What are the pros and cons of `undefined` vs. `null`?

mhuynh5757:
> Left a minor comment. Have you tested reloading the app? Does the array of devices get populated correctly?

Yes, tried with the "test app", reloaded the bundle and invoked "getAudioDevices" multiple times, and reloaded bundle multiple times.

api/voice-react-native-sdk.api.md
mhuynh5757: One thing I can imagine happening is default-valued function parameters. I don't think `null` gets overwritten, but `undefined` does, which I would say is a "pro" for `undefined` over `null`.

CHANGELOG.md
```diff
@@ -28,7 +28,7 @@
 ## Fixes
 
 - Fixed an issue where some types on the `Call` and `Voice` classes were being incorrectly exported. Types and references to `addEventListener` are instead now correctly exported as `addListener`.
-- Fixed an issue where available audio devices were incorrectly emitted and returned by the SDK.
+- Fixed an issue where available audio devices were sometimes incorrectly emitted and returned by the SDK on Android platforms. This occurs more frequently in development environments when the JS bundle is reloaded, but could happen in production environments as well.
```
kbagchiGWC, line 31: @mhuynh5757 - how do you force metro bundler to reload in production environments?

api/voice-react-native-sdk.api.md
charliesantos: It's more of an ease of use related really. When defining a function with params that can be null, it's better to make it optional that way you can call it using the following manner (needs verification)
```ts
foo(bar, baz);

// instead of
foo(bar, baz, null);
```

I understand in your case, it's an event but the mindset just applies the same.

Another reason is, you made it optional [here](https://github.com/twilio/twilio-voice-react-native/pull/131/files#diff-4682b6358cd5eb7f9f7d809200c4cd17864dcf78a845893ee093ff6fe5827e41R13). Just thinking more about consistency.

CHANGELOG.md
mhuynh5757: This statement is more speculation, good point. It's not "forcing Metro bundler to reload" in prod environments, since the JS is already built and bundled, but I'm not sure if that means the native application will never reload the JS bundle.

api/voice-react-native-sdk.api.md
mhuynh5757: I'm pretty settled on making it `undefined` (optional) instead of explicitly `null`, so I'll make the changes. For posterity though, I would like to point out that the `NativeAudioDevicesInfo` interface is what we expect from the native layer, and is separate from our public interfaces. In effect, the SDK will never emit or return that "native" interface.

charliesantos: Gotcha. And sounds good on making it optional.

kbagchiGWC (approved):

charliesantos (approved):

kbagchiGWC (approved):

PR #139, opened 2023-02-22
VBLOCKS-1356 New API for configuring iOS CallKit

bobiechen-twilio:
## Submission Checklist

 - [ ] Tested code changes and observed expected behavior in the example app
 - [ ] Performed a visual inspection of the `Files changed` tab prior to submitting the pull request for review to ensure proper usage of the style guide

## Description

API for iOS CallKit configuration

## Breakdown

- New API for iOS CallKit configuration
- Constants for CallKit configuration options
- Pass options to CallKit initialization

## Validation

- CircleCI


api/voice-react-native-sdk.api.md
```diff
@@ -874,6 +874,7 @@ export interface Voice {
 // @public
 export class Voice extends EventEmitter {
     constructor();
+    configureCallKit(configuration: Record<string, any>): Promise<void>;
```
charliesantos, line 877: - Do we need to configure callkit once for all calls, or once for every call? Just wondering if this should go in the connect method.
- Is it possible to define the `Record`'s possible values instead of just `string, any`? Or are the values dynamic and could change in the future without needing to update the sdk?

bobiechen-twilio: @charliesantos 
- The app can definitely make configuration changes from time to time but the options look more like a one-time setup.
- re: Record - since all the options are optional and have different types, `<string, any>` sounds more suitable for the case, unless there is way to force data type for such case?

charliesantos: - In that case, +1 on making this it's own method. What are your thoughts on renaming this to `setCallKitConfiguration`?
- We can force data types, but whether it makes sense depends on the possible keys and values. What are the possible keys and values for each key? Will these keys and values change dynamically in the future?

bobiechen-twilio: Yes, options can change in the future when Apple finds it necessary. I added supported options in the `@remarks` of the method in `Voice.tsx` but would like you and @mhuynh5757 to help structure the docs so it makes sense for the RN developers.

api/voice-react-native-sdk.api.md
```diff
@@ -917,4 +935,8 @@ export namespace Voice {
     }
 }
 
+// Warnings were encountered during analysis:
+//
+// lib/typescript/type/CallKit.d.ts:15:9 - (ae-forgotten-export) The symbol "Constants" needs to be exported by the entry point index.d.ts
```
mhuynh5757, line 940: @charliesantos not sure what the best solution is here. Basically, I'd like to export the `CallKit.ConfigurationOptions` type but it depends on `Constants` which is the common constants that we use for internal communication between the native and JS layers, so it includes a lot of things we probably don't want to export.

charliesantos: I thought you're able to export them from here? https://github.com/twilio/twilio-voice-react-native/pull/139/files#diff-f5b962e82196241c1d8e9832b56bd5a7dff0e15f2dc9ead4d7a26916c8e957c3R12

mhuynh5757: The issue is that each key is a constant string from the common constants file. For example, `[Constants.CallKitIconTemplateImageData]: string;` and `Constants` is something we are not and should not be exporting.

charliesantos: I see. It's not ideal. Thoughts on actually using the the strings as keys? These are well defined right? Why do we need to put them in the constants file? Why not put them in the ts defs? Example
```ts
  export type ConfigurationOptions = {
    /**
     * Filename of a 80x80 PNG image that will show in the system call UI as the app icon.
     */
    callKitIconTemplateImageData: string;
//...other keys
```

charliesantos: Another option is, if we really need to put them in the constants file, you might need to just stop importing the constants file in `src/type/CallKit.ts` and just "define" the actual configs as suggested above. It's better for customers but you now have two files that needs to be in sync. We can add a comment to each file as a reminder to update the other if we have to change the keys. Or another option is, don't define the config, just like what Bobie already has. But it is not good for customers..

charliesantos (approved):
+1. @bobiechen-twilio please do a quick pass to make sure we didn't miss anything.

ios/TwilioVoiceReactNative.m
```diff
@@ -426,6 +426,14 @@ - (void)sendEventWithName:(NSString *)eventName body:(id)body {
     }
 }
 
+RCT_EXPORT_METHOD(voice_setCallKitConfiguration:(NSDictionary *)configuration
+                  resolver:(RCTPromiseResolveBlock)resolve
+                  rejecter:(RCTPromiseRejectBlock)reject)
+{
+    [self initializeCallKitWithConfiguration:configuration];
+    resolve(@"");
```
mhuynh5757, line 434: Does this resolve `nil` for `Promise<void>`? Sorry, not familiar with the syntax of Obj-C.

mhuynh5757 (approved):

mhuynh5757 (approved):

PR #150, opened 2023-03-08
[VBLOCKS-1486] fix(js): custom parameter type

mhuynh5757:
## Submission Checklist

## Description

This PR changes the custom parameter type to a more restrictive `Record<string, string>` instead of `Record<string, any>`. Note that Android would support `Record<string, boolean | number | string>` but iOS cannot. However, Android will convert `boolean | number | string` into `string` using string representations of the former two, so `Record<string, string>` loses no functionality.

## Breakdown

- Change connect function signature.
- Add changelog entry.

## Validation

- Typescript compiles.

## Additional Notes

Not a lot of testing was done considering it's a type change, only validated by ensuring Typescript compilation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@
 - The integration testing app formerly under `example/` has been renamed/moved to `test/app/`.
 - The React Native dependency within the integration testing app has been updated from `0.63.4` to `0.66.5`.
 
+### API Changes
+- The `voice.connect` method now has the following function signature
+  ```ts
+  voice.connect(token: string, options?: Voice.ConnectOptions);
+
+  interface Voice.ConnectOptions {
+    contactHandle?: string;
+    params?: Record<string, string>;
+  }
+  ```
+  Not passing an options object or leaving any member of the options object undefined will result in those options using default values.
+  See the API documentation for descriptions of options.
+
 ## Features
 
 - The SDK now exports error classes and emits error objects specific to an error code. See the below code snippet for usage.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
   }
   ```
   Not passing an options object or leaving any member of the options object undefined will result in those options using default values.
-  See the API documentation for descriptions of options.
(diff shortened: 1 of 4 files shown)


src/Voice.tsx
```diff
@@ -857,7 +857,7 @@ export namespace Voice {
     /**
      * Custom parameters to send to the TwiML Application.
      */
-    params?: Record<string, any>;
+    params?: Record<string, string>;
```
charliesantos, line 860: No unit tests related to this change?

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@
...
+  ```ts
+  voice.connect(token: string, options?: Voice.ConnectOptions);
+
+  interface Voice.ConnectOptions {
+    contactHandle?: string;
+    params?: Record<string, string>;
+  }
+  ```
+  Not passing an options object or leaving any member of the options object undefined will result in those options using default values.
+  See the API documentation for descriptions of options.
```
charliesantos, line 19: What are the default values? Maybe add a link to the API docs for the `options`

charliesantos (changes requested):

src/Voice.tsx
mhuynh5757: Not sure if we need unit tests for this since there's no logic actually backing this. Either the TS will compile or fail, on our end or a customer's end. WDYT?

CHANGELOG.md
mhuynh5757: Do we have a permanent link to the docs yet? We don't have the Markdown files on `main` yet so not sure what the best way to link this is.

charliesantos: Yes have a permanent link to the docs. It's referenced in the readme
https://github.com/twilio/twilio-voice-react-native/blob/1.0.0-preview.1/docs/voice-react-native-sdk.md
For this PR, just change `1.0.0-preview.1` to `1.0.0-beta.1` so it'll just work once we go beta.

src/Voice.tsx
charliesantos: I think we need some kind of testing. If we accidentally change this API, how do we catch it during build process? Do we have a step in our build process that tests the types?

charliesantos (approved):

PR #182, opened 2023-06-30
DEMO PURPOSES: Example of Custom Notification Design

afalls-twilio:
N/A


kbagchiGWC:
This looks good to me overall. One comment that I added in slack already, what if the developers want to customize only few and not all 5 choices?
```
RemoteViews provideRemoteViews(final String packageName,
                                   final String displayName,
                                   final String contentBanner,
                                   final PendingIntent acceptNotification,
                                   final PendingIntent rejectNotification,
                                   final PendingIntent wakeNotification);
```

PR #209, opened 2023-08-16
VBLOCKS-1677 Move Voice's generic event type definitions to the end.

manjeshbhargav:
## Submission Checklist

### DIFF
--- a/src/Call.tsx
+++ b/src/Call.tsx
@@ -64,29 +64,19 @@ export declare interface Call {
 
   /** @internal */
   emit(
-    callEvent: Call.Event.QualityWarningsChanged,
+    qualityWarningsChangedEvent: Call.Event.QualityWarningsChanged,
     currentQualityWarnings: Call.QualityWarning[],
     previousQualityWarnings: Call.QualityWarning[]
   ): boolean;
+
+  /** @internal */
+  emit(callEvent: Call.Event, ...args: any[]): boolean;
 
   /**
    * ----------------
    * Listener Typings
    * ----------------
    */
-
-  /**
-   * Generic event listener typings.
-   * @param callEvent - The raised event string.
-   * @param listener - A listener function that will be invoked when the event
-   * is raised.
-   * @returns - The call object.
-   */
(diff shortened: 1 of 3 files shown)


mhuynh5757 (changes requested):
The changes for the `Voice` class look good, can we also do the same changes for the `Call` class?

src/Call.tsx
```diff
@@ -64,30 +64,20 @@ export declare interface Call {
...
   emit(
-    callEvent: Call.Event.QualityWarningsChanged,
+    qualityWarningsChangedEvent: Call.Event.QualityWarningsChanged,
     currentQualityWarnings: Call.QualityWarning[],
     previousQualityWarnings: Call.QualityWarning[]
   ): boolean;
 
+  /** @internal */
+  emit(callEvent: Call.Event, ...args: any[]): boolean;
+
```
mhuynh5757, line 74: I'm thinking this one over again, and I think maybe we should remove the `emit(GenericEvent...` types, since the SDK is the one emitting and we never want to emit a generic event, unlike listening for a generic event. What do you think?

manjeshbhargav: @mhuynh5757 Actually, we should not have any emit type definitions at all, since the developer should not be emitting events on this object at all.

mhuynh5757: I took a look at the TSDocs library documentation and it seems like there's no good solution here. The reason we wanted typed `.emit` declarations in the first place is for easier maintenance. If someone on the team tries to do something like `.emit(Call.Event.ConnectFailure)` and forget to pass the argument, then the compiler would fail and let us know we're emitting wrong.

It seems like marking the `.emit` as `@internal` is the proper solution, it signifies that 3rd party developers are not to use the API. But it doesn't hide it entirely, which is one downside. What do you think?

mhuynh5757: Discussed with @manjeshbhargav over Slack, we've come to the conclusion that the `.emit` types should not hurt since they are marked `@internal` and that the `Generic` types also should not hurt since Intellisense will pick up on the other narrower types first.

mhuynh5757 (approved):

PR #233, opened 2023-09-29
VBLOCKS-2093: Improved handling of call invite rejected, one code path now

afalls-twilio:
## Submission Checklist

## Description

Improved such that there exists one code path for the call-rejected messages and verified that indeed, when the activity is running, an event is fired for call-rejected. _**NOTE: Without redesigning the SDK, there is no way to have the JS layer receive the event when the activity is minimized.**_

## Breakdown

- Cleaned up message flow for call-invite-rejected events. This is similar to what was done for call accepted.
- Verified best possible behavior is available.

## Validation

- Ran both test cases, call invite with application foregrounded, and with application backgrounded.

## Additional Notes

None

### DIFF
--- a/android/src/main/java/com/twiliovoicereactnative/Storage.java
+++ b/android/src/main/java/com/twiliovoicereactnative/Storage.java
@@ -27,6 +27,7 @@ public class Storage {
   static final Map<String, CancelledCallInvite> cancelledCallInviteMap = new HashMap<>();
   // A map to keep uuid and js promise objects associated
   static final Map<String, Promise> callAcceptedPromiseMap = new HashMap<>();
+  static final Map<String, Promise> callRejectPromiseMap = new HashMap<>();
 
   static void releaseCallInviteStorage(String uuid, String callSid, int notificationId, String action) {
     Log.d(TAG, "Removing items in callInviteMap" +
@@ -38,5 +39,6 @@ public class Storage {
     Storage.callInviteCallSidUuidMap.remove(callSid);
     Storage.uuidNotificationIdMap.remove(uuid);
     Storage.callAcceptedPromiseMap.remove(uuid);
+    Storage.callRejectPromiseMap.remove(uuid);
   }
 }
--- a/android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
+++ b/android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
@@ -158,10 +158,10 @@ public class VoiceBroadcastReceiver extends BroadcastReceiver {
 
         if (Storage.callRejectPromiseMap.containsKey(uuid)) {
           Promise promise = Storage.callRejectPromiseMap.get(uuid);
-          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "accept");
+          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "reject");
           promise.resolve(uuid);
         } else {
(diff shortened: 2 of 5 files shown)


android/src/main/java/com/twiliovoicereactnative/Storage.java
```diff
@@ -38,5 +39,6 @@ static void releaseCallInviteStorage(String uuid, String callSid, int notificati
     Storage.callInviteCallSidUuidMap.remove(callSid);
     Storage.uuidNotificationIdMap.remove(uuid);
     Storage.callAcceptedPromiseMap.remove(uuid);
+    Storage.callRejectPromiseMap.remove(uuid);
```
mhuynh5757, line 42: Will there be a problem if the `uuid` does not exist as a key in any of these maps? For example, only either `Storage.callAcceptedPromiseMap` or `Storage.callRejectPromiseMap` will actually have a key-value mapping with that `uuid` since those actions are mutually exclusive, right? These methods won't throw or anything?

android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
```diff
@@ -155,6 +155,14 @@ public void onReceive(Context context, Intent intent) {
         params.putMap(EVENT_KEY_CALL_INVITE_INFO, callInviteInfo);
 
         AndroidEventEmitter.getInstance().sendEvent(ScopeVoice, params);
+
+        if (Storage.callRejectPromiseMap.containsKey(uuid)) {
+          Promise promise = Storage.callRejectPromiseMap.get(uuid);
+          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "accept");
```
mhuynh5757, line 161:
```suggestion
          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "reject");
```

android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
```diff
@@ -155,6 +155,14 @@ public void onReceive(Context context, Intent intent) {
...
+          promise.resolve(uuid);
+        } else {
+          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "accept");
```
mhuynh5757, line 164:
```suggestion
          Storage.releaseCallInviteStorage(uuid, callInvite.getCallSid(), Storage.uuidNotificationIdMap.get(uuid), "reject");
```

mhuynh5757 (changes requested):

android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
afalls-twilio: Oh Good catch!

android/src/main/java/com/twiliovoicereactnative/VoiceBroadcastReceiver.java
afalls-twilio: Oh Good catch!

android/src/main/java/com/twiliovoicereactnative/Storage.java
afalls-twilio: No, there wont...
`the previous value associated with key, or null if there was no mapping for key. (A null return can also indicate that the map previously associated null with key.)`

https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#remove-java.lang.Object-

mhuynh5757 (approved):

PR #272, opened 2024-01-09
[VBLOCKS-2464] feat: android call message

kpchoy:
## Submission Checklist

## Description

[VBLOCKS-2464](https://issues.corp.twilio.com/browse/VBLOCKS-2464)

## Breakdown

- Android Call Message Implementation
- Typescript Call Message Implementation

## Validation

- added tests

## Additional Notes


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Twilio Voice React Native SDK has now reached milestone `beta.4`. Included in th
 - Fixed inconsistency with `AudioDevice` typings, preferring `undefined` over `null` for optional values.
 - Fixed an issue with `call.isMuted()` and `call.isOnHold()` APIs. They should now always return `boolean | undefined` instead of potentially returning `null`.
 - Fixed an issue with `call.getFrom()`, `call.getTo()`, and `call.getSid()` APIs. They should now always return `string | undefined` instead of potentially returning `null`.
+- Call Message feature for Android
```
charliesantos, line 16: This is one of the most anticipated features in RN SDK. Let's put it under "Features" section, above "Fixes".

## Features

### Call Message Events (Beta)

// Add new information about the new API and some related links. Also add example.
// You can use this as a reference.

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
```diff
@@ -0,0 +1,74 @@
...
+      )
+    );
+  }
+
+  private void sendJSEvent(@NonNull WritableMap event) {
+    getJSEventEmitter().sendEvent(ScopeCall, event);
+  }
+
+  private void debug(final String message) {
+    logger.debug(String.format(message));
```
afalls-twilio, line 72: String.format doesn't need to be called here.

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
```diff
@@ -396,6 +397,16 @@ public void call_getStats(String uuid,  Promise promise) {
     }
   }
 
+  @ReactMethod
+  public void call_sendMessage(String uuid, CallMessage message, Promise promise) {
+    final CallRecord callRecord = validateCallRecord(reactContext, UUID.fromString(uuid), promise);
+
+    if (null != callRecord) {
```
afalls-twilio, line 404: What is the situation where there isn't a call record? 
* When the message comes via call-invite there should exist a call record
* When the message comes via a call, there should be a call record.

it seems like it should instead have a check to throw an assert if the callRecord is missing
Objects.requireNonNull(...)

afalls-twilio (changes requested):

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
```diff
@@ -396,6 +398,15 @@ public void call_getStats(String uuid,  Promise promise) {
     }
   }
 
+  @ReactMethod
+  public void call_sendMessage(String uuid, CallMessage message, Promise promise) {
```
mhuynh5757, line 402: We can't send whole objects from the JS layer to the native layer, we we will have to construct the `CallMessage` object within the native function. We can only send primitives from the JS layer, i.e. strings in this case. Then, using those strings, we can use the `CallMessage` builder.

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
```diff
@@ -0,0 +1,74 @@
...
+import android.util.Pair;
+
+import androidx.annotation.NonNull;
+
+import com.facebook.react.bridge.WritableMap;
+import com.twilio.voice.Call;
+import com.twilio.voice.CallMessage;
+import com.twilio.voice.VoiceException;
+
+public class CallMessageListenerProxy implements Call.CallMessageListener {
```
mhuynh5757, line 25: There's two more entry points that we'll need to instantiate a `CallMessageListener`, `voice.connect` for outgoing calls, and `callInvite.accept` for incoming calls.

src/CallMessage.tsx
```diff
@@ -0,0 +1,314 @@
...
+ * @public
+ */
+export namespace CallMessage {
+  /**
+   * Enumeration of all event strings emitted by {@link (CallMessage:class)} objects.
+   */
+  export enum Event {
+    /**
+     * Raised when call message fails.
+     * See {@link (CallMessage:interface).(addListener:2)}.
```
mhuynh5757, line 257: The `addListener` numbers should be `1`, `2`, and `3`.

src/Call.tsx
```diff
@@ -806,6 +807,30 @@ export class Call extends EventEmitter {
...
 
+  /**
+   * Send Call Message.
+   *
+   * @example
+   * To send a user-defined-message
+   * const callMessage = {
+   *    content: 'This is a messsage from the parent call',
+   *    messageType: 'user-defined-message',
+   *    contentType: "application/json"
```
mhuynh5757, line 818:
```suggestion
   *    contentType: 'application/json'
```

src/Call.tsx
```diff
@@ -14,6 +14,7 @@ import type {
   NativeCallEventType,
   NativeCallInfo,
 } from './type/Call';
+import type { CallMessageOptions } from './type/CallMessage';
```
mhuynh5757, line 17: Let's move this type to the `CallMessage` namespace in `src/CallMessage.ts`.

src/CallMessage.tsx
mhuynh5757, line 1: This file can actually be `.ts`. The other files are incorrectly using `.tsx`, since none of our source files actually have any JSX in them. If we limit our files to mostly `.ts` and only `.tsx` when we need it, we can clearly see which files will be exporting or using React components.

mhuynh5757: For the sake of stability, we haven't renamed the `.tsx` files to `.ts` yet. We should do that in the future. But I'm ok doing it here since it's a brand new file.

src/Call.tsx
```diff
@@ -806,6 +807,30 @@ export class Call extends EventEmitter {
...
+   *
+   * @param message - A CallMessage object containing content,
+   * contentType, and messageType
+   *
+   * @returns
+   *  A `Promise` that
+   *    - Resolves when the message has been sent.
+   *    - Rejects when the message is unable to be sent.
+   */
+  sendMessage(message: CallMessageOptions): Promise<void> {
```
mhuynh5757, line 830: Let's resolve the `Promise` with the string that the native layer generates.

mhuynh5757 (changes requested):

src/CallMessage.tsx
kpchoy: I noticed you started at 2 for `src/Call.tsx`, is this a typo? 
https://github.com/twilio/twilio-voice-react-native/blob/main/src/Call.tsx#L848

charliesantos (commented):
@kpchoy don't we have new events for when a message is received and when a message is sent?

kpchoy:
> @kpchoy don't we have new events for when a message is received and when a message is sent? EDIT: Looking at the iOS example here, it seems we also have an event for error? https://www.twilio.com/docs/voice/sdks/ios/changelog#670

I have added the documentation in the CHANGELOG for these cases

src/type/Call.ts
```diff
@@ -57,14 +58,22 @@ export interface NativeCallQualityWarningsEvent {
   previousWarnings: NativeCallQualityWarnings;
 }
 
+export interface NativeCallMessageReceivedEvent {
+  type: Constants.CallEventMessageReceived;
+  call: NativeCallInfo;
+  callMessageSID: string;
```
mhuynh5757, line 64: We should reference the common constants for keys of native interfaces, like so:

```
[Constants.CallMessageSID]: string;
```

src/type/CallInvite.ts
```diff
@@ -40,3 +41,13 @@ export interface NativeCancelledCallInviteInfo {
   from: string;
   to: string;
 }
+
+export interface NativeCallInviteMessageReceivedEvent {
+  type: Constants.CallEventMessageReceived;
+  callMessageSID: string;
```
mhuynh5757, line 47: See above.

mhuynh5757 (changes requested):

CHANGELOG.md
```diff
@@ -3,6 +3,36 @@
 
 Twilio Voice React Native SDK has now reached milestone `beta.4`. Included in this version are the following.
 
+## Features
+
+### Call Message Events (Beta)
+
+- Allow for "user-defined" messages during a Voice Call and Call Invite
```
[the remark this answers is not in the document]
kbagchiGWC: during **s/an/a** pending Call Invite

src/CallMessage.tsx
mhuynh5757: Yes, that's an error. We will have to fix that, I will file a ticket.

CHANGELOG.md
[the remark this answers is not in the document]
kpchoy: i dont understand this comment @kbagchiGWC ?

src/CallMessage.ts
```diff
@@ -62,8 +62,19 @@ export declare interface CallMessage {
    */
 
   /**
-   * Failure event. Raised when call message fails.
-   * TODO example
+   * Failure event. Raised when callMessage fails.
```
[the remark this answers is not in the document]
kpchoy: Android says: `Notifies the delegate that a [CallMessage](https://twilio.github.io/twilio-voice-android/docs/latest/com/twilio/voice/CallMessage.html) was failed to be sent out. This feature is currently in Beta.`

https://twilio.github.io/twilio-voice-android/docs/latest/com/twilio/voice/Call.CallMessageListener.html 

Do we want even more description?

android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
```diff
@@ -235,4 +238,16 @@ public static WritableArray serializeCallQualityWarnings(@NonNull Set<Call.CallQ
...
+
+  /**
+   * Serializes a Call Message 
+   * @param CallMessage the call message
+   * @return A serialized Call
+   */
+  public static WritableMap serializeCallMessage(@NonNull final CallMessage callMessage) {
+    return constructJSMap(
+      new Pair<>(CallMessageContent, callMessage.getContent()),
+      new Pair<>(CallMessageContentType, callMessage.getMessageContentType())
```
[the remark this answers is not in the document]
kpchoy: oops missed this 👍

src/type/Call.ts
kpchoy: Do you want to create a ticket to continue the pattern through the rest of the files?

Yes, will remove the outer `callMessageSID`

mhuynh5757: Yeah, here's the ticket: https://issues.corp.twilio.com/browse/VBLOCKS-2643

src/CallMessage.ts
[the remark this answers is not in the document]
mhuynh5757: That's good enough for me TBH.

CHANGELOG.md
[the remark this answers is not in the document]
kbagchiGWC: Was pointing out the grammatical mistake in the commit suggestion

src/OutgoingCallMessage.ts
```diff
@@ -0,0 +1,368 @@
...
+ * - See the {@link (OutgoingCallMessage:namespace).Event} enum for events emitted by `OutgoingCallMessage`
+ *   objects.
+ * - See the {@link (OutgoingCallMessage:interface) | OutgoingCallMessage interface} for overloaded event listening
+ *   metods.
+ * - See the {@link (OutgoingCallMessage:namespace) | OutgoingCallMessage namespace} for types and enumerations
+ *   used by this class.
+ *
+ * @public
+ */
+export class OutgoingCallMessage extends EventEmitter implements CallMessage {
```
charliesantos, line 154: Our original purpose for extending CallMessage is to avoid duplication of code. In this implementation, we are duplicating all of CallMessage members. I understand we also need to inherit from EventEmitter and multiple inheritance is tricky in typescript as it doesn't support it natively. There are multiple ways to address this but I'm hesitant to implement something complex for this simple use case. I'm ok if we compromise for something like: CallMessage inherits from EventEmitter. That way, OutgoingCallMessage can inherit CallMessage properly without duplication of code. Downside is, CallMessage is now an EventEmitter. I don't see any negative side effects though. Thoughts @mhuynh5757 @kpchoy ?

src/__tests__/Call.test.ts
```diff
@@ -525,6 +540,45 @@ describe('Call class', () => {
         await expect(postFeedbackPromise).resolves.toBe(undefined);
       });
     });
+
+    describe('.sendMesssage', () => {
```
charliesantos, line 544: Can we add negative test cases?

src/__tests__/CallInvite.test.ts
```diff
@@ -251,6 +261,108 @@ describe('CallInvite class', () => {
       expect(to).toBe(createNativeCallInviteInfo().to);
     });
   });
+
+  describe('.sendMessage()', () => {
```
charliesantos, line 265: Please add negative test cases

src/CallMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Provides access to information about a callMessage, including the call
+ * message content, contentType, messageType, and voiceEventSid
+ *
+ * @remarks
+ * Note that the callMessage information is fetched as soon as possible from the
+ * native layer, but there is no guarantee that all information is immediately
+ * available. Methods such as `CallMessage.getContent` or `CallMessage.getSid`
+ * may return `undefined`.
```
charliesantos, line 20: In what scenario where these return undefined? How can a customer know when these are available?

charliesantos (changes requested):

api/voice-react-native-sdk.api.md
```diff
@@ -297,6 +324,36 @@ export namespace CallKit {
...
+    // @internal
+    constructor({ content, contentType, messageType, voiceEventSid, }: NativeCallMessageInfo);
+    _content: string;
+    _contentType: CallMessage.ContentType;
+    getContent(): string;
+    getContentType(): CallMessage.ContentType;
+    getMessageType(): CallMessage.MessageType;
+    getSid(): string | undefined;
+    _messageType: CallMessage.MessageType;
+    _voiceEventSid?: string;
```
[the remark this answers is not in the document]
kpchoy: These members are public, because I have `OutgoingCallMessage` implement `CallMessage`, if I make them private I get type error.

https://github.com/twilio/twilio-voice-react-native/blob/VBLOCKS-2464/feat-call-message/src/OutgoingCallMessage.ts#L154

api/voice-react-native-sdk.api.md
```diff
@@ -544,6 +601,48 @@ namespace MediaErrors {
...
+export class OutgoingCallMessage extends EventEmitter implements CallMessage {
+    constructor({ content, contentType, messageType, voiceEventSid, }: NativeCallMessageInfo);
+    _content: string;
+    _contentType: CallMessage.ContentType;
+    getContent(): string;
+    getContentType(): CallMessage.ContentType;
+    getMessageType(): CallMessage.MessageType;
+    getSid(): string | undefined;
+    _messageType: CallMessage.MessageType;
+    _voiceEventSid?: string;
```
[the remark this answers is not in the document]
kpchoy: These members are public, because I have `OutgoingCallMessage` implement `CallMessage`, if I make them private I get type error.

https://github.com/twilio/twilio-voice-react-native/blob/VBLOCKS-2464/feat-call-message/src/OutgoingCallMessage.ts#L154

src/OutgoingCallMessage.ts
kpchoy: I think @mhuynh5757 was concerned that a User might add a listener to `CallMessage` even though it doesn't have any listeners configured. I do see issue with duplicate code. @mhuynh5757 did you have any other points?

src/CallMessage.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Message type
+   */
+  public _messageType: CallMessage.MessageType;
+
+  /**
+   * Unique identifier of the message
+   */
+  public _voiceEventSid?: string;
```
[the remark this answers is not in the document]
kpchoy: ref: https://github.com/twilio/twilio-voice-react-native/pull/272#discussion_r1468050819

src/OutgoingCallMessage.ts
mhuynh5757: I don't think Typescript mixins are overly complex, should we give it a shot and see how it looks? @kpchoy what do you think? I don't think it will take more than half an hour of experimentation but it might be throwaway work if we decide it's not worth it.

src/CallMessage.ts
kpchoy: We have similar notes in `Call.ts` for getters. https://github.com/twilio/twilio-voice-react-native/blob/VBLOCKS-2464/feat-call-message/src/Call.tsx#L335

I will remove unless @mhuynh5757 objects?

api/voice-react-native-sdk.api.md
```diff
@@ -297,6 +324,36 @@ export namespace CallKit {
...
+export namespace CallMessage {
+    // (undocumented)
+    export enum ContentType {
+        // (undocumented)
+        'ApplicationJson' = "application/json"
+    }
+    // (undocumented)
+    export enum MessageType {
+        // (undocumented)
+        'UserDefinedMessage' = "userDefinedMessage"
```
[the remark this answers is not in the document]
mhuynh5757: It does not look like it gets sent to the server like this.

src/CallMessage.ts
charliesantos: The concern is, if customers see this, they would likely have the same question.

src/OutgoingCallMessage.ts
charliesantos: I've seen instances where mixins become unnecessarily complex but if it takes only this much time to experiment, feel free to go ahead.

charliesantos: Also, mixins is only one option. Feel free to do a bit of exploration and choose what you think fits our needs.

api/voice-react-native-sdk.api.md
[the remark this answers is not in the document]
kpchoy: We send this to the Android layer, then the java layer matches the string and sends actual type to server: https://github.com/twilio/twilio-voice-react-native/blob/VBLOCKS-2464/feat-call-message/android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java#L542 

I am fine changing the name to `user-defined-message` wdyt @charliesantos ?

src/OutgoingCallMessage.ts
kpchoy: Ok Ill give mixins a shot

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
```diff
@@ -45,8 +44,7 @@ public void onMessageSent(String voiceEventSID) {
     // notify JS layer
     sendJSEvent(
       constructJSMap(
-        new Pair<>(VoiceEventType, CallEventMessageSent),
-        new Pair<>(CallMessageSID, voiceEventSID)
+        new Pair<>(VoiceEventType, CallEventMessageSent)
```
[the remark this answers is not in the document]
kpchoy: Do we still need the Voice Event SID here, if we don't emit it on the JS layer?

ex:
```
   * outgoingCallMessage.addListener(OutgoingCallMessage.Event.Sent, () => {
   *    // outgoingCallMessage sent
   * })
```

api/voice-react-native-sdk.api.md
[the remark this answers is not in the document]
charliesantos: We should change it to `user-defined-message`. It's an actual type of message type that is consistent across JS, Android, iOS and our backend.

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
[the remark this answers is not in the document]
mhuynh5757: Yes, we still need it. If we do not have the voice event SID, we would not be able to associate with the respective `CallMessage` object on the JS layer.

src/CallMessage.ts
mhuynh5757: It's different for the `Call` objects since there are situations (i.e. outgoing calls) where sometimes values are `undefined` for lengths of time. However, since we construct the `CallMessage` objects ourselves for outgoing call messages, and we'll always have the values defined, we don't need `undefined` here.

src/OutgoingCallMessage.ts
kpchoy: After review of mixins and some experimentation I think it is too complex for our current implementation. I changed it back to the original way Charlie suggested: `CallMessage inherits from EventEmitter. That way, OutgoingCallMessage can inherit CallMessage properly without duplication of code.`. Although it is still a possibility for customers to mistakenly add listeners to CallMessage, I think the possibility is slim.

src/__tests__/Call.test.ts
kpchoy: I could only think of 1 negative case for this method. Take a look, and let me know if you have another case I should implement.

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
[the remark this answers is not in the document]
kpchoy: To associate the `CallMessage` object in the JS layer I did `voiceEventSid` validation here: https://github.com/twilio/twilio-voice-react-native/pull/272/files#diff-09b5abc597ac9e668697ffd940810409846eb726e0aefe5b9ea144fd35c6ab20R225 

Let me know what you think

src/OutgoingCallMessage.ts
charliesantos: I was leaning more towards to this solution as well. I'm not too concerned about customers adding listeners to CallMessage because we are not documenting any events in the public API docs. Thoughts @mhuynh5757 ?

src/__tests__/Call.test.ts
charliesantos: Can you link the commit here? Sorry I should've written down what I was thinking in the first review. I'm thinking cases like the ones we're checking for here https://github.com/twilio/twilio-voice.js/blob/master/lib/twilio/call.ts#L904
- send empty params
- send missing properties etc

charliesantos (commented):
Super close! I left one more comment re: test cases.
Also, have you verified this manually?

src/OutgoingCallMessage.ts
mhuynh5757: Agreed. I think we should just do the `CallMessage extends EventEmitter` route. I tried some shenanigans with typing out the inherited methods on just the root `CallMessage` class and using `never` but it's not worth it.

mhuynh5757: It won't be the end of the world if developers add useless listeners to the root `CallMessage` object.

src/__tests__/Call.test.ts
kpchoy: commit to throw error if param is not `CallMessage`. https://github.com/twilio/twilio-voice-react-native/pull/272/commits/e1a7ea775cfa530d019741d106026dd230db2ab3 

Let me try adding those other 2 cases

kpchoy: @charliesantos additional negative cases: https://github.com/twilio/twilio-voice-react-native/pull/272/commits/61036c0927a2393845084a51f0b606af27e56cf8

kpchoy:
> Super close! I left one more comment re: test cases. Also, have you verified this manually?

How would we verify manually? I thought we would need to implement in example app to verify it works?

charliesantos:
> > Super close! I left one more comment re: test cases. Also, have you verified this manually?
> 
> How would we verify manually? I thought we would need to implement in example app to verify it works?

I was thinking you'd create quick/small test code (nothing fancy) to verify that your implementation actually works. It's ok if we don't have that yet. But we need to submit this PR to a feature branch instead, to make sure we're not merging any untested feature on main.

src/Call.tsx
```diff
@@ -896,21 +897,41 @@ export class Call extends EventEmitter {
...
+    }
+
+    if (
+      typeof message.getContent() === 'undefined' ||
+      message.getContent() === null
+    ) {
+      throw new InvalidArgumentError('"content" is empty');
+    }
+
+    //@ts-ignore
```
mhuynh5757, line 913: Why do we need to `ts-ignore` here?

src/CallInvite.tsx
```diff
@@ -370,21 +371,41 @@ export class CallInvite extends EventEmitter {
...
+      throw new InvalidArgumentError('"content" is empty');
+    }
+
+    //@ts-ignore
+    if (message.getMessageType().length === 0) {
+      throw new InvalidArgumentError(
+        '"messageType" must be a non-empty string'
+      );
+    }
+
```
mhuynh5757, line 393: What do you think about moving this validation logic to a reusable function? Even a method on the `CallMessage` object perhaps?

src/CallMessage.ts
```diff
@@ -13,34 +14,31 @@ import type { NativeCallMessageInfo } from './type/CallMessage';
...
   /**
-   * Content body of the message
+   * The content of the message which should match the contentType parameter.
    */
-  public _content: string;
+  private _content: any;
 
   /**
-   * MIME type for the message
+   * The MIME type of the content. application/json and is the only contentType
```
mhuynh5757, line 26: I think we should leave the description as `The MIME type of the content.` and leave the rest out. We can document what is supported in the `CallMessage.ContentType` enum.

src/OutgoingCallMessage.ts
```diff
@@ -243,9 +222,11 @@ export class OutgoingCallMessage extends EventEmitter implements CallMessage {
       );
     }
 
-    const { message, code } = nativeCallMessageEvent.error;
-    const error = constructTwilioError(message, code);
-    this.emit(OutgoingCallMessage.Event.Failure, error);
+    if (this.getSid() === nativeCallMessageEvent.voiceEventSid) {
```
mhuynh5757, line 225: You should do this check on line `208` before you invoke the handler at all. That way you can cover this for every handler without having to redo the check in every handler.

src/OutgoingCallMessage.ts
```diff
@@ -262,46 +243,10 @@ export class OutgoingCallMessage extends EventEmitter implements CallMessage {
       );
     }
 
-    this.emit(OutgoingCallMessage.Event.Sent);
+    if (this.getSid() === nativeCallMessageEvent.voiceEventSid) {
```
mhuynh5757, line 246: See above.

mhuynh5757 (changes requested):

android/src/main/java/com/twiliovoicereactnative/CallMessageListenerProxy.java
```diff
@@ -0,0 +1,84 @@
...
+        new Pair<>(JSEventKeyCallMessageInfo, serializeCallMessage(callMessage))
+      )
+    );
+  }
+
+  private void sendJSEvent(@NonNull WritableMap event) {
+    getJSEventEmitter().sendEvent(ScopeCallMessage, event);
+  }
+
+  private void debug(final String message) {
```
afalls-twilio, line 81: No need for this function, just call the logger directly

afalls-twilio (changes requested):
Where are you registering to receive call messages? Look at the test cases in android voice CI as an [example](https://github.com/twilio/twilio-voice-android/blob/d0acac2edf953886d42d85d93f0e32c33207ba35/voice/src/androidTest/java/com/twilio/voice/EventingTest.java#L425)

src/Call.tsx
kpchoy: oops left over from previous implementation, removed 👍

kpchoy: It's because this case is type protected. `Property 'length' does not exist on type 'MessageType'.`. Since `MessageType` is an enum. Should I remove this error case?

kpchoy:
> Where are you registering to receive call messages? Look at the test cases in android voice CI as an [example](https://github.com/twilio/twilio-voice-android/blob/d0acac2edf953886d42d85d93f0e32c33207ba35/voice/src/androidTest/java/com/twilio/voice/EventingTest.java#L425)

Will register once I write the integration tests in separate PR

src/Call.tsx
kpchoy: done

also 5.xx upgrade ticket: https://issues.corp.twilio.com/browse/VBLOCKS-2673

kpchoy:
Also created feature branch `feature/user-defined-messages` @charliesantos

src/Call.tsx
```diff
@@ -897,25 +896,7 @@ export class Call extends EventEmitter {
...
-      throw new InvalidArgumentError('"content" is empty');
-    }
-
-    //@ts-ignore
-    if (message.getMessageType().length === 0) {
-      throw new InvalidArgumentError(
-        '"messageType" must be a non-empty string'
-      );
-    }
+    verifySendMessage(message);
```
mhuynh5757, line 899: Instead of executing this check when we want to send the message, should we execute it when constructing the message object?

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ const message = new CallMessage({
    contentType: CallMessage.ContentType.ApplicationJson,
    messageType: CallMessage.MessageType.UserDefinedMessage
 })
-const outgoingCallMessage: OutgoingCallMessage = await call.sendMessage(message)
+const outgoingCallMessage: OutgoingCallMessage = await call.sendMessage(message: CallMessage)
```
mhuynh5757, line 18:
```suggestion
const outgoingCallMessage: OutgoingCallMessage = await call.sendMessage(message);
```

mhuynh5757 (changes requested):

src/Call.tsx
kpchoy: This follows the same pattern as the js-sdk https://github.com/twilio/twilio-voice.js/blob/master/lib/twilio/call.ts#L904 

We can, but what if user doesn't send a message? 

@charliesantos any reason for doing the checks before sending message instead of when constructing the message object?

CHANGELOG.md
kpchoy: sorry, should reflect this comment now: https://github.com/twilio/twilio-voice-react-native/pull/272#discussion_r1472148568

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
```diff
@@ -523,6 +536,21 @@ Call.Issue getIssueFromString(String issue) {
...
+
+  CallMessage.MessageType getMessageTypeFromString (String messageType) {
+    if (messageType.equals(CommonConstants.UserDefinedMessage)) {
+      return CallMessage.MessageType.UserDefinedMessage;
+    }
+    return CallMessage.MessageType.UserDefinedMessage;
+  }
+
+  String getContentTypeFromString (String contentType) {
+    if (contentType.equals(CommonConstants.ApplicationJson)) {
```
afalls-twilio, line 548: same as above

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
```diff
@@ -523,6 +536,21 @@ Call.Issue getIssueFromString(String issue) {
     }
     return Call.Issue.NOT_REPORTED;
   }
+
+  CallMessage.MessageType getMessageTypeFromString (String messageType) {
```
afalls-twilio, line 540: Is this method private? is it being called anywhere outside the class? if not, lets make it private. Also maybe change the name to validate message type.. also, instead of silently failing, should we report some error to the user, maybe a js exception or something?

afalls-twilio (changes requested):

src/Call.tsx
charliesantos: It's ok either way. As long as you perform validation before sending it to the server.

mhuynh5757: I strongly prefer doing the check upon construction, that way the developer knows as soon as possible the `CallMessage` is invalid and cannot be sent.

kpchoy: done 👍

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
kpchoy: done made private 👍 

@mhuynh5757 I see the method above this `getIssueFromString` defaults return value, so I followed similar pattern. Any opinions on throwing an error to user?

mhuynh5757: Hmm, let's see. I can see it being useful, but a bit more work in the long run. These functions don't really do validation, more like type guarding. We already ensure that the passed values are enumeration members in the JS layer, so it's not strictly necessary to throw here, but it would reduce errors like typos on our end down the road. I think we should do it, but it will also create a bit more work when we have new enumeration members. When we support more types, we'll have to modify both the native and JS layer. But that seems OK to me, what do you think?

kpchoy: I have added the error to Java layer. It can't hurt to add more validation where we can.

afalls-twilio (approved):
LGTM

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
mhuynh5757: Very nice, let's do one more thing though. If we look at the control flow here, we're returning `null` to the builder which will probably cause the Java runtime to crash. For example, let's say message type is invalid and we return `null`. The Java runtime is still executing, and passes `null` to the builder, which will cause an exception. So we're going to want to do these validations earlier, the promise will be rejected properly, and then we want to return instead of throwing.

mhuynh5757 (changes requested):

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
```diff
@@ -404,8 +404,11 @@ public void call_sendMessage(String uuid, String content, String contentType, St
     final CallRecord callRecord = validateCallRecord(reactContext, UUID.fromString(uuid), promise);
     Objects.requireNonNull(callRecord);
 
-    final CallMessage callMessage = new CallMessage.Builder(validateMessageTypeFromString(messageType, promise))
-      .contentType(validateContentTypeFromString(contentType, promise)).content(content).build();
+    final CallMessage.MessageType _messageType = validateMessageTypeFromString(messageType, promise);
+    final String _contentType = validateContentTypeFromString(contentType, promise);
+
```
mhuynh5757, line 409: Sorry, just one more thing. We're not actually modifying the control flow here yet. What happens with this control flow when `messageType` is invalid?

mhuynh5757 (changes requested):

android/src/main/java/com/twiliovoicereactnative/TwilioVoiceReactNativeModule.java
kpchoy: added if statements to just return if validations are null

mhuynh5757 (approved):

PR #386, opened 2024-07-18
[VBLOCKS-2965] Call message e2e tests

mhuynh5757:
## Submission Checklist

## Description

This PR implements e2e tests for the call message feature in the RN SDK.

## Breakdown

- Add relay server for interacting with the Twilio Call Message feature.
- Add tests that use the relay server.

## Validation

- Locally ran tests.

## Additional Notes

N/A


.circleci/config.yml
```diff
@@ -308,6 +308,13 @@ jobs:
             cd test/app/android/app
             echo $GOOGLE_SERVICES_JSON > google-services.json
             cd ../../../..
+      - run:
+          name: "Start relay server"
+          background: true
+          command: |
+            cd test/app
+            yarn run detox:relay-server
+            cd ../..
```
kpchoy, line 317: Do we need to do this? I thought circleci defaults to working_directory with every run?

test/app/e2e/relay/server.js
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * @type {CreateSubscription}
+   */
+  const createSubscription = (callSid, callbackUrl) => {
+    const url =
+      `${extendedUrl}/${callSid}/UserDefinedMessageSubscriptions.json`;
+
+    const params = new URLSearchParams();
+    params.append('Method', 'POST');
```
kpchoy, line 76: is this `POST` needed if you already using `post` on line 79?

test/app/e2e/relay/server.js
```diff
@@ -0,0 +1,212 @@
...
+
+    res.json(subscription.data);
+  });
+
+  app.post('/receive-message', (req, res) => {
+    const isValidTwilioRequest = twilio.validateExpressRequest(req, authToken);
+    if (!isValidTwilioRequest) {
+      res.sendStatus(400);
+      console.warn('Incoming request is not from Twliio');
+      return;
```
kpchoy, line 141: since you throw error at 149, does it make sense to throw error here as well instead of just returning?

.circleci/config.yml
mhuynh5757: The working directory would be the root of the project, not the nested `test/app` folder. I found that I needed this in the past, but things might have changed. The rest of the config file uses this pattern.

test/app/e2e/relay/server.js
mhuynh5757: We need it in both locations, yes. The Twilio API defines the protocol as such. The first POST (in params append) is actually part of the data we're sending to Twilio. The second POST is the actual POST request we're making.

test/app/e2e/relay/server.js
mhuynh5757: I opted not to throw on the first "error" since we don't want the server to crash if we receive a request from a non-Twilio source. It will simply just ignore the request and give the sender a 400. However, I opted to throw (and therefore crash) the server at the second "error" since that means Twilio sent us something wrong, and that would indicate a deeper issue in the tests.

.circleci/config.yml
kpchoy: Using `cd ../..` will take you back to the root, but I think the default is to start at root with each run so no need to go back. But yeah there is no harm in leaving this.

kpchoy (approved):

PR #438, opened 2024-10-16
feat: proof of concept appium testing

mhuynh5757:
## Submission Checklist

## Description

This PR adds a proof of concept test to the repo test app using appium.

## Breakdown

- Add appium automated testing to the test app.

## Validation

- Locally ran appium tests.

## Additional Notes

N/A

### DIFF
--- a/test/app/src/e2e/OutgoingCall.tsx
+++ b/test/app/src/e2e/OutgoingCall.tsx
@@ -1,0 +1,85 @@
+import * as React from 'react';
+import { Button, Text, View } from 'react-native';
+import { Call, Voice } from '@twilio/voice-react-native-sdk';
+import { generateAccessToken } from '../tokenUtility';
+
+type TestStatus = 'not-started' | 'running' | 'success' | 'failed';
+
+type TestHook = () => { status: TestStatus; run: () => Promise<void> };
+
+async function settlePromise<T>(
+  p: Promise<T>
+): Promise<{ result: 'ok' | 'err'; value: T }> {
+  return p
+    .then((value) => ({ result: 'ok' as const, value }))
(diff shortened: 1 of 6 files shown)


test/app/src/e2e/OutgoingCall.tsx
```diff
@@ -0,0 +1,85 @@
...
+      connectFailurePromise,
+    ]);
+
+    if (callStatus === 'connectFailure') {
+      setStatus('failed');
+      return;
+    }
+
+    await new Promise((resolve) => {
+      setTimeout(resolve, 5000);
```
kpchoy, line 55: is the manual wait time still necessary? appium doesn't have a way to `.wait`?

kpchoy (commented):
Do you have an example I can look at for how the other team is using appium?
1. So we would need to create the UI for every test we make?
2. Since this is just 1 test case and has ~100 lines, each test case would be around the same length?
3. One concern I see is that it looks tedious to manually set status for each test.

test/app/src/e2e/OutgoingCall.tsx
mhuynh5757: So appium does have a way to wait, but the way I set up the test such that each test is it's own UI and each hook is it's own test run time means that appium isn't in control here. At this point, it's app code, and the appium wait is more similar to how jest waits for tests to finish since appium is the test runner.

mhuynh5757:
Yeah those concerns are very valid. I have a few ideas on how to cut down on boilerplate, but in the end it will be pretty lengthy for each test compared to the type of testing we're used to. If you compare to how the current tests work, with a "full blown" test app and just doing e2e testing on that, it's not going to be too different in terms of total length.

kpchoy (approved):

PR #699, opened 2026-07-15
Revert "[VBLOCKS-6682] feat: ignore accept and reject for invalid callinvites"

mhuynh5757:
Reverts twilio/twilio-voice-react-native#695

Instead of merging to main while we restore CI, let's merge to `ci-interim` branch. A follow-up PR will be opened to merge this feature work to `ci-interim` after reverting these changes merged to the `main` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,3 @@
-2.0.0-preview.3 (In Progress)
-=============================
-
-## Fixes
-
-### Platform Specific Fixes
-
-#### Android
-
-- Fixed null pointer exception related crashes that could occur when accepting or rejecting invalid CallInvites using the native notification.
-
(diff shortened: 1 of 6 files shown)


mhuynh5757:
Choosing to ignore the Copilot comments for now considering we're reverting a bugfix so that we can put it on a work-in-progress branch instead of main.

charliesantos:
How about we update ci-interim branch from main? Why do we need to revert this?

mhuynh5757:
@charliesantos main was updated with the Android changes but no automated CI was actually run. That PR got merged recently in June. I felt like it was not a good idea to leave changes in main that were not CI-tested, so this PR fixes that mistake I made.

charliesantos (approved):

PR #700, opened 2026-07-16
[VBLOCKS-4141] Add unknown type audio device

mhuynh5757:
## Submission Checklist

## Description

This PR implements a new AudioDevice type member with value "unknown". This is a fallback audio device type for when the OS returns an unexpected audio device type. Now, there is also a `nativeType` member of the AudioDevice that can be used to see the value that the native layer returns.

This solution was chosen instead of further marshaling because the underlying OS and those values are subject to continuous change.

## Breakdown

- Add "unknown" to the AudioDevice enum.
- Add new "nativeType" to the AudioDevice type.

## Validation

- Tested on iOS with a Bluetooth device that previously was detected as BluetoothA2DP, which would fail the enum and be passed in audioDevice.type and not fall under the JS enum value set. Now, the reported audio device is `{ type: AudioDevice.Type.Unknown, nativeType: 'BluetoothA2DP' }`.
- Tested on a real Android device in a real outgoing call by mocking the returned type from the native layer. Mocked to "unknown" and the audio device selection still works properly, as well as the `audioDevice.type` and `audioDevice.nativeType` are proper.

## Additional Notes

N/A

### DIFF
--- a/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
+++ b/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
@@ -2,8 +2,10 @@ package com.twiliovoicereactnative;
 
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyAudioDevices;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyName;
+import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyNativeType;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeySelectedDevice;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyType;
+import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyUnknown;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyUuid;
 import static com.twiliovoicereactnative.CommonConstants.CallInfoFrom;
 import static com.twiliovoicereactnative.CommonConstants.CallInfoInitialConnectedTimestamp;
@@ -177,11 +179,12 @@ class ReactNativeArgumentsSerializer {
    */
   public static WritableMap serializeAudioDevice(String uuid, @Nullable AudioDevice audioDevice) {
     if (null != audioDevice) {
-      String type = audioDevice.getClass().getSimpleName();
+      String nativeType = audioDevice.getClass().getSimpleName();
       return constructJSMap(
         new Pair<>(AudioDeviceKeyUuid, uuid),
         new Pair<>(AudioDeviceKeyName, audioDevice.getName()),
-        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.AUDIO_DEVICE_TYPE.get(type)));
+        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.AUDIO_DEVICE_TYPE.getOrDefault(nativeType, AudioDeviceKeyUnknown)),
+        new Pair<>(AudioDeviceKeyNativeType, nativeType));
     }
     return null;
   }
--- a/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
+++ b/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
@@ -5,7 +5,6 @@ import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyName;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyNativeType;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeySelectedDevice;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyType;
-import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyUnknown;
 import static com.twiliovoicereactnative.CommonConstants.AudioDeviceKeyUuid;
 import static com.twiliovoicereactnative.CommonConstants.CallInfoFrom;
 import static com.twiliovoicereactnative.CommonConstants.CallInfoInitialConnectedTimestamp;
@@ -183,7 +182,7 @@ class ReactNativeArgumentsSerializer {
       return constructJSMap(
         new Pair<>(AudioDeviceKeyUuid, uuid),
         new Pair<>(AudioDeviceKeyName, audioDevice.getName()),
-        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.AUDIO_DEVICE_TYPE.getOrDefault(nativeType, AudioDeviceKeyUnknown)),
+        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.getAudioDeviceType(audioDevice)),
         new Pair<>(AudioDeviceKeyNativeType, nativeType));
     }
     return null;
--- a/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
+++ b/android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
@@ -178,7 +178,7 @@ class ReactNativeArgumentsSerializer {
    */
   public static WritableMap serializeAudioDevice(String uuid, @Nullable AudioDevice audioDevice) {
     if (null != audioDevice) {
-      String nativeType = audioDevice.getClass().getSimpleName();
+      String nativeType = AudioSwitchManager.getAudioDeviceNativeType(audioDevice);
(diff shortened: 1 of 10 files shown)


src/type/AudioDevice.ts
```diff
@@ -5,6 +5,7 @@ import type { Uuid } from './common';
 export interface NativeAudioDeviceInfo {
   uuid: Uuid;
   type: AudioDevice.Type;
+  nativeType: string;
   name: string;
 }
```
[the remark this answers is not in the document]
mhuynh5757: I don't believe this applies here, this project requires Expo development builds and the native code is included in this PR.

android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
```diff
@@ -177,11 +179,12 @@ public static WritableMap serializeCall(@NonNull final CallRecord callRecord) {
...
   public static WritableMap serializeAudioDevice(String uuid, @Nullable AudioDevice audioDevice) {
     if (null != audioDevice) {
-      String type = audioDevice.getClass().getSimpleName();
+      String nativeType = audioDevice.getClass().getSimpleName();
       return constructJSMap(
         new Pair<>(AudioDeviceKeyUuid, uuid),
         new Pair<>(AudioDeviceKeyName, audioDevice.getName()),
-        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.AUDIO_DEVICE_TYPE.get(type)));
+        new Pair<>(AudioDeviceKeyType, AudioSwitchManager.AUDIO_DEVICE_TYPE.getOrDefault(nativeType, AudioDeviceKeyUnknown)),
+        new Pair<>(AudioDeviceKeyNativeType, nativeType));
```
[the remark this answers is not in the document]
mhuynh5757: This value has been stable between all builds.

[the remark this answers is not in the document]
mhuynh5757: Investigating a solution now...

mhuynh5757:
@charliesantos pinging you early for a review so we can discuss the breaking change this PR will introduce. What do you think about the proposed changes? I think they're safe enough for the 2.x release. The only break will be exhaustiveness on the enumeration so maybe a few users might see compilation errors if they have exhaustive checks.

charliesantos:
> @charliesantos pinging you early for a review so we can discuss the breaking change this PR will introduce. What do you think about the proposed changes? I think they're safe enough for the 2.x release. The only break will be exhaustiveness on the enumeration so maybe a few users might see compilation errors if they have exhaustive checks.

Before I review, does this PR includes all the required changes?

mhuynh5757:
@charliesantos yeah the PR contains all the necessary changes. I just finished testing, I thought I found a bug with audio device selection not working but it was a bug in the Test App and not the SDK code.

I pinged early so we could discuss the breaking changes and I was open to feedback to change anything if need be.

But in any case, it's actually complete now and fully tested and I'll move the PR into non-draft mode.

android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
```diff
@@ -177,11 +178,12 @@ public static WritableMap serializeCall(@NonNull final CallRecord callRecord) {
    */
   public static WritableMap serializeAudioDevice(String uuid, @Nullable AudioDevice audioDevice) {
     if (null != audioDevice) {
-      String type = audioDevice.getClass().getSimpleName();
+      String nativeType = audioDevice.getClass().getSimpleName();
```
charliesantos: Building on @Copilot's point - `getClass().getSimpleName()` returns the R8/ProGuard-minified name (e.g. `'a'`) in a consuming app's release build, so `nativeType` becomes useless exactly where it is meant to serve as a diagnostic escape hatch. This is the same fragility the PR already fixed in `getAudioDeviceType()` by switching to `instanceof`. Define stable string constants in `CommonConstants` and populate `nativeType` via the same `instanceof` chain used there.

ios/TwilioVoiceReactNative.m
```diff
@@ -268,7 +272,7 @@ - (NSString *)audioPortTypeMapping:(NSString *)portType {
         return kTwilioVoiceReactNativeAudioDeviceKeyBluetooth;
     }
 
-    return portType;
+    return kTwilioVoiceReactNativeAudioDeviceKeyUnknown;
 }
 
 - (BOOL)selectAudioDevice:(NSString *)uuid {
```
charliesantos, line 278: For an unknown-type device neither the earpiece nor the bluetooth branch matches, so `portDescription` stays nil and `setPreferredInput:nil` clears the input, returning YES without routing to the selected device. Add an else branch that matches `device[kTwilioVoiceAudioDeviceUid]` against `availableInputs` (same pattern as the bluetooth path), and return NO if no port matches.

src/__tests__/AudioDevice.test.ts
```diff
@@ -58,5 +66,9 @@ describe('AudioDevice namespace', () => {
       expect(AudioDevice.Type).toBeDefined();
       expect(typeof AudioDevice.Type).toBe('object');
     });
+
```
charliesantos, line 77: Consider adding a round-trip test that constructs an AudioDevice with `type: AudioDevice.Type.Unknown` and asserts `audioDevice.type === AudioDevice.Type.Unknown` - the current test only checks the constant value, not the construction path that is the point of this PR.

src/AudioDevice.tsx
```diff
@@ -33,9 +33,29 @@ export class AudioDevice {
...
 
+  /**
+   * The type of the audio device as reported verbatim by the native layer, without
+   * being mapped to a {@link (AudioDevice:namespace).Type}.
+   *
+   * @remarks
+   * Native platform APIs can report audio device types that are not represented by
+   * {@link (AudioDevice:namespace).Type}, for example due to OS updates introducing
+   * new device categories. This field is provided as an escape hatch for those cases
+   * and its value is not guaranteed to be stable across native SDK or OS versions.
```
charliesantos, line 55: Worth tweaking the caveat to distinguish platforms: on iOS `nativeType` is the Apple-defined `AVAudioSessionPort` constant (OS-stable), while on Android it is the AudioSwitch class identifier (library-version-dependent, and fragile under R8 until the serializer finding is resolved).

src/__mocks__/AudioDevice.ts
```diff
@@ -18,22 +19,26 @@ export function createNativeAudioDevicesInfo(): NativeAudioDevicesInfo {
       {
         uuid: 'mock-nativeaudiodeviceinfo-uuid-one',
         type: 'earpiece' as AudioDevice.Type,
+        nativeType: 'mock-nativeaudiodeviceinfo-nativetype-one',
```
charliesantos, line 22: Worth considering a fixture with a realistic Unknown-type `nativeType` (e.g. `'BluetoothA2DP'`) to exercise the fallback path - the current generic placeholders are fine for round-trip tests but not representative. Low priority.

charliesantos (changes requested):
This targets the 2.0 preview branch so the breaking change is acceptable, but the required `nativeType` field and the new `Type.Unknown` enum member should be called out under a Breaking Changes heading in the CHANGELOG, not only under Features/Fixes. See inline for the Android R8 stability gap on `nativeType`. (3 findings from the local review were omitted as already raised by other reviewers.)

android/src/main/java/com/twiliovoicereactnative/ReactNativeArgumentsSerializer.java
mhuynh5757: ~That change was already pushed, please take a look again.~

My comment is wrong, I misread the above. Sorry about that.

ios/TwilioVoiceReactNative.m
mhuynh5757: Using an `else-if` statement as opposed to a `else` statement so as to not disturb the Speaker fallthrough case.

src/__mocks__/AudioDevice.ts
mhuynh5757: The fallback related code is all in the native layer. Adding a JS unit test does not exercise any different code paths.

mhuynh5757: Added a reference to audio device testing to the existing appium testing ticket we have in the backlog VBLOCKS-6582

charliesantos (changes requested):
@mhuynh5757 I think you missed this part
>...the required nativeType field and the new Type.Unknown enum member should be called out under a Breaking Changes heading in the CHANGELOG, not only under Features/Fixes.

mhuynh5757:
@charliesantos you're right sorry about that 🤦

charliesantos (approved):

bobiechen-twilio (approved):
syntax looks good to me

PR #702, opened 2026-07-24
fix(ios): report VoIP pushes synchronously instead of on a timer (PRO-5725)

jhutchins:
## Summary

Fixes the recurring `-[PKPushRegistry _terminateAppIfThereAreUnhandledVoIPPushes]` SIGABRT on iOS, seen in production builds shipped *after* PRO-4264's fix (`1.7.0-cc.2`, merged 2026-06-12). Same crash signature, new root cause — the previous fix's premise doesn't hold.

Linear: PRO-5725

## Root cause

PRO-4264 assumed both `callInviteReceived:` and `cancelledCallInviteReceived:` resolve **asynchronously** (a "network round-trip"), so it deferred any unresolved VoIP push completion to a 3-second `dispatch_after` watchdog.

That assumption is only half right, confirmed via Twilio's own SDK docs and changelog:
- `callInviteReceived:` fires **synchronously**, on the same queue, inside `handleNotification:`.
- `cancelledCallInviteReceived:error:` fires **asynchronously** — "if any of the following events occur: the call is prematurely disconnected by the caller, or the callee does not accept or reject the call in ~30s."

More importantly, per Apple DTS (developer.apple.com/forums/thread/801446) and matching real-world crash telemetry ("`callservicesd`: ... error: Killing VoIP app because it failed to post an incoming call **in time**"):

> When you return from `didReceiveIncomingPushWith`, PKPushRegistry checks whether `reportNewIncomingCall` was made. If not, the app is terminated. There is no grace period — it is not a timer.

So for a cancel-type push (or any payload Twilio's SDK silently drops), `didReceiveIncomingPushWithPayload:...` returns having reported nothing, and the 3s watchdog — a deferred timer — cannot reliably win a race against an OS check that fires the moment the delegate method returns.

I confirmed this against the actual crash log (`.ips`) attached to PRO-5725: the crashing frame is `voipPayloadReceived:mustPostCall:withCompletionHandler:_block_invoke`, dispatched via `_dispatch_main_queue_drain`, invoking `_terminateAppIfThereAreUnhandledVoIPPushes` — consistent with PushKit's own deferred main-queue audit finding no report for that push, not with any race inside our own code (all relevant work here is single-threaded on the main queue; ruled that out explicitly).

I also confirmed there is **no drift** in `ios/TwilioVoicePushRegistry.m` between the `1.7.0-cc.2` tag and `main` — the file investigated here is exactly what shipped, so this is a genuinely new gap in the previously-fixed code, not a regression from an unrelated change.

## Fix

Remove the deferred 3-second watchdog. Immediately after `handleNotification:` returns, if `sPendingVoIPCompletion` is still set (i.e. nothing resolved synchronously — a cancel push or a dropped payload), satisfy it with a placeholder report **right there**, before `didReceiveIncomingPushWithPayload:...` returns. This guarantees the "report before return" contract unconditionally, with zero window for the race, instead of hoping a timer beats an undocumented OS deadline.

As a result, the `sVoIPPushGeneration` counter (which existed solely to let the watchdog detect it had gone stale) is now dead code and removed along with it.

## Also considered and ruled out

- **Double `reportNewIncomingCall` for the same UUID / duplicate-report throw** — traced through `TVPRSatisfyPendingPushWithPlaceholder` and `callInviteReceived:`; the shared `sPendingVoIPCompletion` static is always nil'd out before use, so no code path can invoke a given push's completion more than once. Not evidenced by the crash log either (no ObjC frames from our code are on the aborting stack).
- **Real invite resolving *after* its cancel was already placeholder-satisfied** (the previously flagged "residual") — this specific mechanism requires `callInviteReceived:` to resolve asynchronously, which Twilio's docs say it does not. Not reproducible for genuine invite pushes.
- **Cross-thread data race on the static vars** — ruled out; `PKPushRegistry` uses `dispatch_get_main_queue()`, and Twilio's `delegateQueue:nil` also defaults to main per SDK docs, so `didReceiveIncomingPushWithPayload:`, `callInviteReceived:`, `cancelledCallInviteReceived:`, and the (now-removed) watchdog all serialize on the same queue.

## Test plan

- [ ] Native iOS build (no Xcode/CocoaPods environment available in this sandbox to run a full build; changes are syntax/brace-balance verified and reuse only APIs already present elsewhere in the file — no new symbols introduced)
- [ ] Manual device verification of: normal invite+answer, invite+decline, invite+timeout (30s), fast invite/cancel pair, standalone cancel push with no prior invite delivered to this device
- [ ] Confirm crash-free session rate for `_terminateAppIfThereAreUnhandledVoIPPushes` after rollout

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jhutchins:
Opened against the wrong repo by mistake — this is an internal fix specific to our fork's push-registry architecture, not intended for upstream. Closing.

PR #703, opened 2026-07-28
[VBLOCKS-6966] GHA release pipeline

mhuynh5757:
## Submission Checklist

## Description

This PR implements a release pipeline to replace our CircleCI release process with GHA.

## Breakdown

- GHA artifactory, CI, and release pipeline configurations.

## Validation

- Manual testing in separate test repo.

## Additional Notes

N/A


scripts/npm-dist-tag.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+function getNpmDistTag(version) {
+  if (/-rc[0-9]+$/.test(version)) {
+    return 'rc';
+  }
+
+  const match = version.match(
+    /^[0-9]+\.[0-9]+\.[0-9]+-([a-z]+)\.?[0-9]+(?:-.*)?$/
+  );
+  return match ? match[1] : 'latest';
```
luisrivas: `getNpmDistTag('2.0.0-dev')` returns latest, that's the version currently in `package.json`. Same for `1.0.0-alpha` or any typo'd prerelease identifier without a numeric suffix. Please add `2.0.0-dev` and `1.0.0-alpha` test cases to `npm-dist-tag.test.js`.

github-temp/workflows/ci.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Artifactory OIDC Auth
+        uses: ./.github/actions/artifactory-oidc
+        with:
+          artifactory_url: ${{ vars.ARTIFACTORY_URL }}
+
+      # Copied to a plain filename since upload-artifact needs an unambiguous path.
+      - name: Stage .npmrc for hand-off to test
+        run: cp ~/.npmrc ./artifactory.npmrc
+
+      - name: Upload .npmrc
```
kpchoy: **[high] Artifactory token exposed via downloadable artifact on a public repo**

The token is uploaded here as an artifact; on a public repo, run artifacts are downloadable by anyone while the token is live. SECURITY.md's "not rendered anywhere" reasoning conflates not-logged with not-accessible.

**Suggestion:** Re-run the OIDC exchange per job, or hand off via a masked `$GITHUB_OUTPUT` instead of an artifact.

github-temp/workflows/prepare-release.yml
```diff
@@ -0,0 +1,193 @@
...
+        run: npm version --no-git-tag-version "${RELEASE_VERSION}"
+
+      - name: Stamp constants files
+        env:
+          RELEASE_VERSION: ${{ inputs.release_version }}
+        run: node ./scripts/substitute-constants-version.js "${RELEASE_VERSION}"
+
+      - name: Build SDK artifacts
+        run: yarn run build:docs
+
```
kpchoy, line 73: **[high] Release build output is never verified**

`if-no-files-found` defaults to `warn`, and downstream `git add -f` of an empty path adds nothing without failing — a no-op build ships as a release.

**Suggestion:** Set `if-no-files-found: error` and assert `lib/` non-empty + constants stamped before committing.

scripts/npm-dist-tag.js
kpchoy: **[medium] Typo/unrecognized prerelease falls through to `latest`**

`return match ? match[1] : 'latest'` sends any unrecognized shape (a typo'd manual `release_version`) to the `latest` dist-tag and force-moves the git `latest` tag.

**Suggestion:** `throw` on unrecognized prerelease shapes so a bad dispatch fails loudly.

github-temp/workflows/ci.yml
mhuynh5757: We don't want to run the OIDC handshake per job because that requires `id-token: write` capability. Supposedly the artifacts are secure but I will re-run my research on that to make sure I'm not missing anything.

scripts/npm-dist-tag.js
mhuynh5757: This was actually an intentional decision on my part when I made this script, since we never publish `1.0.0-beta`, we would actually publish `1.0.0-beta.0`. But if what you suggest is actually more accurate to proper semver, I would agree with you and we should add those changes.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,70 @@
...
+          exit 1
+        fi
+        echo "::add-mask::$ART_TOKEN"
+
+        # 3. Point npm/yarn at the curated Artifactory registry for this job.
+        REGISTRY="${BASE_URL}/artifactory/api/npm/virtual-npm-thirdparty/"
+        {
+          echo "registry=${REGISTRY}"
+          echo "//${HOST}/artifactory/api/npm/virtual-npm-thirdparty/:_authToken=${ART_TOKEN}"
+        } > ~/.npmrc
```
luisrivas, line 106: This action only configures `~/.npmrc`, which is used by others steps; however, Yarn Berry reads `.yarnrc.yml` instead of `~/.npmrc`.

luisrivas (changes requested):

github-temp/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
+name: Release
```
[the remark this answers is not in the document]
mhuynh5757: Known, not an issue. We want to discuss and merge temp files before having GitHub pick up on these Actions.

github-temp/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+
+env:
+  ARTIFACTORY_URL: ${{ vars.ARTIFACTORY_URL }}
+
+jobs:
+  # id-token: write isn't audience-scoped - a job holding it can request an
+  # OIDC token for any audience, not just Artifactory's. Isolated here so no
+  # dependency-installing job ever holds this permission.
+  artifactory-auth:
+    runs-on: ubuntu-x64
```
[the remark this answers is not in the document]
mhuynh5757: Internal runner name. Non-issue.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,70 @@
...
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
+          -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${BASE_URL}" \
+          | jq -r '.value')
```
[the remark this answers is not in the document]
mhuynh5757: Non-issue for now. Will revisit if this becomes an issue, but this action was given to us by the platform team, so minimal changes are desired.

github-temp/workflows/ci.yml
mhuynh5757: Did research and revisited this. Will add the `id-token: write` to each job. We should be safe via the `production` environment requirement anyways.

github-temp/workflows/prepare-release.yml
mhuynh5757: Good idea. I think we can revisit this when we address future things in the ADR such as build provenance attestation.

scripts/npm-dist-tag.js
mhuynh5757: Good suggestion. I think we can hone the regex down the line if we encounter any issues with it. I'll add a note.

scripts/npm-dist-tag.js
```diff
@@ -0,0 +1,44 @@
...
+    return 'rc';
+  }
+
+  // Consider stricter regex in the future to prevent unexpected prerelease
+  // tags making it through this check.
+  const match = version.match(
+    /^[0-9]+\.[0-9]+\.[0-9]+-([a-z]+)\.?[0-9]+(?:-.*)?$/
+  );
+  return match ? match[1] : 'latest';
+}
```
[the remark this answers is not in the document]
mhuynh5757: Doesn't apply to current version and tags that we historically have used. Will revisit if need-be.

scripts/npm-dist-tag.js
```diff
@@ -0,0 +1,44 @@
...
+  if (/-rc[0-9]+$/.test(version)) {
+    return 'rc';
+  }
+
+  // Consider stricter regex in the future to prevent unexpected prerelease
+  // tags making it through this check.
+  const match = version.match(
+    /^[0-9]+\.[0-9]+\.[0-9]+-([a-z]+)\.?[0-9]+(?:-.*)?$/
+  );
+  return match ? match[1] : 'latest';
```
kpchoy: **[critical] Version classification is duplicated, inconsistent, and silently defaults to `latest`**

This function and `scripts/is-release-candidate.js` each parse the version with their own regex, and they disagree on real inputs (verified empirically):
- `1.0.0-rc.1` (what `npm version prerelease --preid rc` emits): `getNpmDistTag` returns `rc`, but `isReleaseCandidate` returns `false`. Since `release.yml` only skips publishing when `IS_RELEASE_CANDIDATE == 'true'`, a normally-built RC is published instead of held.
- `1.0.0-beta`, `1.0.0-dev`, `1.0.0-RC1`, `v1.0.0-beta.1`, and `""` all fall through the `[a-z]+\.?[0-9]+` regex to `return 'latest'` — a prerelease or empty/mis-resolved version string gets published to the `latest` dist-tag, overwriting the real release, with no error.

**Suggestion:** Derive both the dist-tag and the RC flag from a single classifier so they cannot diverge (make `isReleaseCandidate === (distTag === 'rc')`), and throw on anything unclassifiable instead of defaulting to `latest`. Keep it dependency-free — `npm-dist-tag.js` runs in the `deploy` job, which by design has no `node_modules`, so `require('semver')` is unavailable there; the official semver regex works without a dependency.

scripts/is-release-candidate.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Given a semver version string, returns whether it's a release candidate
+ * (bare X.Y.Z-rcN, or compound X.Y.Z-preview.N-rcM/X.Y.Z-betaN-rcM).
+ * Detected by checking the version's last hyphen-delimited segment.
+ */
+function isReleaseCandidate(version) {
+  const lastSegment = version.slice(version.lastIndexOf('-') + 1);
+  return /^rc[0-9]+$/.test(lastSegment);
```
kpchoy: **[critical] RC detection diverges from dist-tag logic (see npm-dist-tag.js)**

`/^rc[0-9]+$/` on the last hyphen segment rejects the dotted form `rc.1`, while `npm-dist-tag.js` accepts it — so the publish gate and the chosen dist-tag are computed from divergent logic for canonical RC versions like `1.0.0-rc.1`. Fold this into the shared classifier referenced in the `npm-dist-tag.js` comment.

**Suggestion:** Delete this regex; return `classifyVersion(version).isReleaseCandidate` from the same source of truth.

github-temp/workflows/release.yml
```diff
@@ -0,0 +1,115 @@
...
+          fi
+
+      - name: Determine npm dist-tag and release-candidate status
+        id: dist-tag
+        run: |
+          VERSION="$(jq -r '.version' package.json)"
+          echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+
+          RELEASE_TAG=$(node ./scripts/npm-dist-tag.js "$VERSION")
```
kpchoy: **[critical] Publish decision consumes unvalidated script output (fail-open gate)**

`RELEASE_TAG=$(node ./scripts/npm-dist-tag.js …)` and `IS_RELEASE_CANDIDATE=$(node …)` drive the two most consequential decisions with no validation. The gate `if: env.IS_RELEASE_CANDIDATE != 'true'` is fail-open: if the script prints nothing (or a value other than `true`), `'' != 'true'` is true and it publishes anyway. An empty `RELEASE_TAG` feeds `npm publish --tag ""`.

**Suggestion:** Allowlist-check both captures before publishing:
```bash
case "$RELEASE_TAG" in latest|rc|beta|preview|alpha|next) ;; *) echo "::error::unexpected dist-tag '$RELEASE_TAG'"; exit 1;; esac
[ "$IS_RELEASE_CANDIDATE" = true ] || [ "$IS_RELEASE_CANDIDATE" = false ] || { echo "::error::bad rc flag"; exit 1; }
```

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,80 @@
...
+        BASE_URL=$(echo "${ARTIFACTORY_URL}" | sed -E 's#/+$##')
+        if [ -z "$BASE_URL" ]; then
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
```
kpchoy: **[high] `curl` lacks `--fail`; `GITHUB_JWT` becomes the literal string `"null"`**

`GITHUB_JWT=$(curl -sS … | jq -r '.value')` — on an HTTP 4xx/5xx (e.g. missing `id-token: write`), `curl -sS` still exits 0 and `jq -r '.value'` prints `null`. `set -euo pipefail` does not catch it (both processes exit 0). Note the asymmetry: the exchange step below correctly guards with `.access_token // empty`, but this step has no guard, so `"null"` flows into `subject_token` and surfaces later as a misleading "token exchange failed."

**Suggestion:** `curl -sS --fail-with-body` on both calls, `jq -er '.value // empty'`, and an explicit `[ -n "$GITHUB_JWT" ]` check before the exchange.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,80 @@
...
+
+        # 4. Same for yarn, which ignores .npmrc from v2 onward and needs its
+        #    own config. Written to the home directory, not the project: a
+        #    project .yarnrc.yml is usually checked in (nodeLinker etc.) and
+        #    overwriting it would change how the project installs.
+        {
+          echo "npmRegistryServer: \"${REGISTRY}\""
+          echo "npmAlwaysAuth: true"
+          echo "npmAuthToken: \"${ART_TOKEN}\""
+        } > ~/.yarnrc.yml
```
kpchoy, line 116: **[high] Artifactory token written to `$HOME` with no cleanup, on custom runners**

The token is written to `~/.npmrc` and `~/.yarnrc.yml` with no `post`/`always` cleanup. All jobs use `runs-on: ubuntu-x64` — a custom label, not a GitHub-hosted one, implying self-hosted/ARC. `SECURITY.md`'s "credential never leaves the job" reasoning silently assumes ephemeral runners; nothing enforces it. If runners are reused, the live token persists on disk for the next job. Compounded by `ci.yml` then running six `yarn install`s with lifecycle scripts enabled while that token is on disk — a compromised transitive dep could read and exfiltrate it.

**Suggestion:** Write auth config to a `$RUNNER_TEMP` path (via `--userconfig` / `YARN_RC_FILENAME`) or add an `always()` cleanup step; confirm and document that `ubuntu-x64` runners are ephemeral.

kpchoy (changes requested):
Solid pipeline with a thorough `SECURITY.md`, but there are real issues: see the 5 inline comments

scripts/npm-dist-tag.js
luisrivas: I see. I have one more question: release_version is a free-text dispatch input, so what `dist-tag` would `2.x.y-dev` get here?

mhuynh5757: @luisrivas I've updated the implementation. See the new unit tests for clarity on this behavior.

The new implementation would throw an error when trying to get the `npm` release tag or the `isReleaseCandidacy` of `2.0.0-dev`.

scripts/npm-dist-tag.js
mhuynh5757: Reimplemented. Thanks for the feedback!

scripts/is-release-candidate.js
mhuynh5757: Reimplemented, thanks for the feedback!

github-temp/workflows/release.yml
mhuynh5757: Reimplemented, thanks for the feedback! There's no allowlist checking in the shell scripting of the release pipeline but the new version parsing is airtight enough that I don't think this is an issue.

.github/actions/artifactory-oidc/action.yml
mhuynh5757: There's a common platform-hosted Artifactory OIDC handshake action repo that we will use instead of the vendorized one formerly given to us to copy-paste verbatim. I'd like to hold off on improving the OIDC handshake action since we'll be migrating to that one in the next PR.

github-temp/workflows/release.yml
```diff
@@ -0,0 +1,139 @@
...
+          artifactory_url: ${{ vars.ARTIFACTORY_URL }}
+
+      - run: corepack enable
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # 7.0.0
+        with:
+          node-version: "22.13"
+          cache: yarn
+
+      - run: yarn install --immutable
```
kpchoy: The `test` and `deploy` jobs both have `id-token: write` and both live in `release.yml`. Because npm's OIDC workflow claim is per-file, a token minted in `test` and a real publish token from `deploy` look identical to npm on the repo and workflow claims. `test` also runs `yarn install`, which executes dependency lifecycle scripts, so a compromised transitive dependency's postinstall could mint a token here and publish to npm directly, skipping `deploy`'s approval gate, tag check, and `--ignore-scripts`.

The only thing distinguishing the two is the `production` environment claim, which `deploy` has and `test` does not. That protects us only if npm's trusted publisher is configured to require `environment: production`. That is an off-repo setting on npmjs.com, and if it is left blank, publishes still succeed and nothing fails, so the control can silently go missing.

This commit was already tested by `ci.yml` pre-merge, so the re-test here is largely redundant. Dropping the `test` job, or moving it to its own workflow file so its token carries a different workflow claim (the way `build` in prepare-release.yml already does), takes dependency-code execution out of the publish file and makes the environment setting a backup again instead of the only barrier. SECURITY.md already flags this ("Dropping this job would harden that"); worth deciding whether to close it here or track it as a follow-up.

github-temp/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Enable Corepack so `yarn` resolves to the version pinned in package.json's
+      # "packageManager" field (Yarn 4) instead of the runner's global Yarn 1.
+      # Must run before setup-node, whose `cache: yarn` step invokes yarn.
+      - run: corepack enable
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # 7.0.0
+        with:
+          node-version: "22.13"
+          cache: yarn
```
kpchoy, line 40: Nit (non-blocking): `cache: yarn` keys only on the root `yarn.lock`, but the job also installs four sub-project workspaces (`test/app`, `test/expo`, `test/appium-harness`, `test/appium-orchestrator`). Set `cache-dependency-path` to cover `test/*/yarn.lock` so their dependency changes invalidate the cache. Cache-effectiveness only, not correctness.

github-temp/workflows/release.yml
mhuynh5757: Good thing to bring it, it was partly why the initial implementation had the Artifactory OIDC handshake as a separate job which would forward it to the `yarn install` job later.

Platform team is responsible for the trusted publisher to require `environment: production` and this is indeed something that protects us in this case.

The `test` job in the `release.yml` file is indeed redundant, I wanted to keep it as a sanity check but maybe it's just best to drop it.

github-temp/workflows/ci.yml
mhuynh5757: This stuff will likely change when we get e2e testing working, will address in the future.

kpchoy (approved):
Great work! This PR is a tough one

luisrivas (approved):
Nice work, @mhuynh5757 !

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,80 @@
...
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
+          -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${BASE_URL}" \
+          | jq -r '.value')
+        echo "::add-mask::$GITHUB_JWT"
```
charliesantos, line 50: Guard `GITHUB_JWT` before use. `jq -r '.value'` yields the literal string `null` on a non-200 or missing `.value`, so `::add-mask::null` masks the word everywhere and the exchange sends `subject_token:"null"` - a confusing JFrog 400 instead of a clear failure at the source. After the mask line add: `if [ -z "$GITHUB_JWT" ] || [ "$GITHUB_JWT" = 'null' ]; then echo "::error::Failed to obtain GitHub OIDC JWT"; exit 1; fi`.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,80 @@
...
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
+          -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${BASE_URL}" \
```
charliesantos: URL-encode the audience. `BASE_URL` is appended raw as `&audience=https://...`, and `:`/`/` are RFC 3986 reserved in a query value; curl won't encode them and strict IdP validators reject it. Use `curl -G --data-urlencode "audience=${BASE_URL}" ...`. (Extends @Copilot's point.)

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,80 @@
...
+          | jq -r '.value')
+        echo "::add-mask::$GITHUB_JWT"
+
+        # 2. Exchange it for an Artifactory access token. The body MUST be JSON
+        #    and MUST include provider_name, or JFrog issues no token.
+        RESP=$(curl -sS "${BASE_URL}/access/api/v1/oidc/token" \
+          -H 'Content-Type: application/json' \
+          -d "{\"grant_type\":\"urn:ietf:params:oauth:grant-type:token-exchange\",
+               \"subject_token_type\":\"urn:ietf:params:oauth:token-type:id_token\",
+               \"subject_token\":\"${GITHUB_JWT}\",
```
charliesantos: Build the token-exchange body with `jq`, not string interpolation. `"${GITHUB_JWT}"` is spliced into a JSON literal unescaped; a malformed value containing `"` or `\` breaks the JSON or injects fields. Use `jq -n --arg tok "$GITHUB_JWT" --arg prov github-actions '{grant_type:"...",subject_token_type:"...",subject_token:$tok,provider_name:$prov}'`.

github-temp/workflows/prepare-release.yml
```diff
@@ -0,0 +1,180 @@
...
+      # Consider checking build artifacts here: VBLOCKS-7002
+
+      - name: Upload release tree
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # 7.0.1
```
charliesantos, line 103: An empty build tree still ships a release. If `build:docs` produces nothing, `upload-artifact` still succeeds (the glob includes always-present `package.json`), `create-release-branch` force-adds nothing, and a release branch ships without built output. `if-no-files-found: error` does NOT catch this - the always-present files satisfy the check. Add an explicit `test -d lib && test -n "$(ls -A lib)"` (and the same for `docs/api/`) as a step before upload, failing the job. (Extends @kpchoy.)

github-temp/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+          TAG="${GITHUB_REF#refs/tags/}"
+          PKG_VERSION="$(jq -r '.version' package.json)"
+          if [ "$TAG" != "$PKG_VERSION" ]; then
+            echo "::error::tag $TAG != package.json version $PKG_VERSION"
+            exit 1
+          fi
+
+      - name: Read version and release-candidate status
+        run: |
+          VERSION="$(jq -r '.version' package.json)"
```
charliesantos, line 113: This `run:` block has no `set -e`, so if `classify-version-tag.js getIsReleaseCandidate` exits non-zero (unrecognized version) `IS_RELEASE_CANDIDATE` is never written - both the publish and skip steps then evaluate false and the job exits success without publishing or erroring. Consider adding `set -euo pipefail` to the block.

charliesantos (changes requested):
Soundly designed for a staged PR - the workflows are inactive under `github-temp/`, so nothing here goes live until the follow-up `.github/` cutover. The SECURITY.md is a genuinely strong artifact and the credential-isolation reasoning holds up.

One cross-cutting note that doesn't anchor cleanly to one line: `latest` can regress if releases approve out of order (SECURITY.md known item) - if 1.7.0's `deploy` finishes after 1.7.1's, both the npm dist-tag and the git tag move backward with no error. Add a compare against the currently-published `latest` before moving either.

github-temp-dispatch/DECISIONS.md
```diff
@@ -0,0 +1,208 @@
+# Release pipeline: the dispatch flow
```
charliesantos, line 1: Question: are you planning to keep DECISIONS.md in this repo long-term, or fold it into the VnV eng-docs repo? It reads like durable design rationale that belongs alongside the other CI/CD proposals in voice-video-sdks-eng-docs rather than living next to the workflows - what's the plan once the dispatch flow moves to `.github/`?

github-temp-dispatch/workflows/release.yml
```diff
@@ -0,0 +1,197 @@
...
+        default: false
+      development_version:
+        description: "Optional: next dev version to commit onto the release branch (e.g. 2.0.1-dev). Leave blank to skip."
+        required: false
+
+# Commented out for now, consider adding later when we have a chance to debug
+# why this causes inconsistent pipeline starts.
+# # Only allow one release in flight at a time. Subsequent releases are queued
+# # rather than cancelled.
+# concurrency:
```
charliesantos: Question: should we re-enable the concurrency guard before this goes live? With it commented out, two release dispatches close together race on the npm `latest` dist-tag and the `latest` git tag (last-write-wins). DECISIONS.md accepts this as operator-paced, but for a publish pipeline it seems worth restoring once the inconsistent-start issue is debugged - what's the thinking on shipping without it?

charliesantos (commented):
Re-reviewed at the dispatch-flow head - the earlier findings are all addressed (JWT guard, audience encoding, jq-built exchange body, the build-tree verify step). Two questions before this lands.

github-temp-dispatch/DECISIONS.md
mhuynh5757: Original plan was to delete it after it answered all the PR questions. Outstanding items would have been made into JIRA tickets and addressed down the road. But I do like the idea of putting it in the VNV-Docs repo if that's what we want. WDYT?

github-temp-dispatch/workflows/release.yml
mhuynh5757: Thinking of filing a JIRA for this one and trying to work on it next sprint. I know that some of the JS projects are blocked on this pipeline so I punted it for now.

FWIW when I was testing in my private repo, when I uncommented this block, it was still when I was on the publish hook. When I published, it would _sometimes_ fire the pipeline, sometimes not. And even sometimes fire it like 30 minutes later. So I felt it was more hassle than worth.

github-temp-dispatch/DECISIONS.md
charliesantos: Thanks for the additional context. I suggest we put this in the vnvdocs repo. I think there are similar documents already regarding build pipelines for your reference.

github-temp-dispatch/workflows/release.yml
charliesantos: Thoughts on this @luisrivas @kpchoy ?
If it's not too much work, let's address it in this PR. I'm ok to address it in the next sprint if the work is significant.

mhuynh5757: @charliesantos as long as we're ok with blocking the JS projects for a bit longer, I'll get started now while the other two weigh in.

kpchoy: I agree if its low effort (~100 lines) lets add it to this PR

github-temp-dispatch/DECISIONS.md
mhuynh5757: Ok, I will do that. Here's my idea:

- Merge this PR if the team is onboard with the implementation in `github-temp-dispatch/`
- Make the next PR to remove the `github-temp/`, leftover markdown docs, and other miscellaneous 
  - This PR will also rename the workflows from `github-temp-dispatch/` to `.github/workflows/...` and make them live.
- Separate PR in VNVDocs that has the DECISIONS.md file appropriately adjusted for that repo's style

github-temp-dispatch/workflows/release.yml
mhuynh5757: @charliesantos @kpchoy @luisrivas I was able to test and get concurrency working. Was simpler than I expected!

kpchoy (commented):
Really nice rewrite. The dispatch flow is a much better shape than inferring intent from the Release checkbox, and the concurrency block looks good now. I re-checked last round's items and they're all properly fixed: the version classifier is a single source of truth that throws instead of defaulting, the JWT null guard and jq-built exchange body are in, and every dispatch input reaches shell through `env:` rather than `${{ }}`.

Three things below, all on the publish path, all variations on the same theme: a step that fails by doing nothing and still reports green. The first is the one I'd want closed before this goes live.

github-temp-dispatch/workflows/release.yml
```diff
@@ -0,0 +1,195 @@
...
+
+          # Down the road, guard this against dispatching from the wrong ref:
+          # require GITHUB_REF to match refs/heads/release/* before publishing.
+          # Dispatching from main would otherwise publish main's -dev version,
+          # which has no built artifacts committed.
+
+          VERSION="$(jq -r '.version' package.json)"
+          echo "::notice::Publishing ${VERSION} under the ${NPM_DIST_TAG} dist-tag."
+
+          npm publish --registry https://registry.npmjs.org \
```
kpchoy, line 78: **`deploy` publishes whatever is checked out, with no verification that it was built**

`lib/`, `src/constants.ts` and `docs/api/` are gitignored, so they exist only on a branch where `prepare-release` committed them with `git add -f`. `prepare-release.yml` has a "Verify the release tree was built" step; there's no equivalent between the artifact commit and the irreversible publish.

Dispatched from any other ref, `npm publish` doesn't fail. `files` globs that match nothing are silently skipped by npm's packlist, and `--ignore-scripts` disables the `prepare` script that would otherwise rebuild them. The result is a green publish of a tarball where `main: lib/commonjs/index` and `types: lib/typescript/index.d.ts` don't exist, under whichever dist-tag was selected, possibly `latest`.

This is worse than "dispatched from the wrong branch" reads: it isn't publishing the wrong version, it's publishing a package that can't be imported. Since `deploy` already carries the comment marking where the ref guard goes, I'd land both here:

```bash
case "$GITHUB_REF" in
  refs/heads/release/*)
    ;;
  *)
    echo "::error::release.yml must be dispatched from a release/* branch"
    exit 1
    ;;
esac

for ENTRY_POINT in "$(jq -r .main package.json)" "$(jq -r .types package.json)"; do
  if [ ! -s "$ENTRY_POINT" ]; then
    echo "::error::${ENTRY_POINT} is missing; refusing to publish"
    exit 1
  fi
done
```

github-temp-dispatch/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Commit release bump
+        env:
+          RELEASE_VERSION: ${{ inputs.release_version }}
+        run: |
+          set -euo pipefail
+
+          git checkout -b "release/${RELEASE_VERSION}"
+          git add package.json
+          git add constants/constants.src
+          git add -f docs/api
```
kpchoy, line 133: **`create-release-branch` never verifies the tree it commits**

The "Verify the release tree was built" step ran in a different job on a different runner, so it says nothing about what `download-artifact` actually restored here.

`git add -f lib` on an existing-but-empty directory exits 0 and stages nothing, and the commit still succeeds because `git add package.json` always stages the version bump. `if-no-files-found: error` on the upload doesn't close it either, since it fires only when no pattern matched anything and `package.json` always matches.

Cheapest fix is asserting the index after staging:

```bash
for RELEASE_DIRECTORY in lib docs/api; do
  if ! git diff --cached --name-only | grep -q "^${RELEASE_DIRECTORY}/"; then
    echo "::error::no ${RELEASE_DIRECTORY}/ files staged for the release commit"
    exit 1
  fi
done
```

github-temp-dispatch/workflows/release.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Commit development-version continuation
+        env:
+          DEVELOPMENT_VERSION: ${{ inputs.development_version }}
+        run: |
+          set -euo pipefail
+
+          npm version --no-git-tag-version "${DEVELOPMENT_VERSION}"
+          node ./scripts/substitute-constants-version.js "${DEVELOPMENT_VERSION}"
+          git rm -r --ignore-unmatch \
```
kpchoy, line 183: **`git rm -r --ignore-unmatch` exits 0 on every unmatched path**

This is the only thing keeping the release-branch merge to a clean version-bump diff. `--ignore-unmatch` is defined to succeed when a pathspec matches nothing, so if any of these five paths is renamed or relocated the removal silently no-ops. `set -euo pipefail` can't help, the command genuinely succeeded, and the commit still lands because `git add -A` staged the version bump.

Worth asserting the post-condition, or dropping `--ignore-unmatch` and listing only paths guaranteed present:

```bash
for RELEASE_PATH in docs/api lib src/constants.ts; do
  if git ls-files --error-unmatch "$RELEASE_PATH" >/dev/null 2>&1; then
    echo "::error::${RELEASE_PATH} is still tracked after git rm"
    exit 1
  fi
done
```

Related, and the reason this one matters more than it looks: the removal lives only inside `continue-development`, gated on `inputs.development_version != ''`. Leaving that optional field blank is documented as a supported choice ("Leave blank to skip"), and taking it means the artifacts stay on the branch. The flow then has a human merge `release/<version>` into `main` on the strength of the artifacts netting out, and `main` ends up tracking `lib/**`, `docs/api/**` and `src/constants.ts`, all gitignored. Either make the input required with a default, or move the cleanup into its own job gated on the same upstream results minus the `development_version` check.

charliesantos (approved):

mhuynh5757:
@kpchoy good feedback. I think we can address all of the three concerns about build artifacts and validation (as well as the validation of proper removal of build artifacts in the dev-continuation commit) in the VBLOCKS-7002 ticket.

kpchoy (approved):

PR #707, opened 2026-08-11
[VBLOCKS-6968] ICE options callinvite accept

mhuynh5757:
## Submission Checklist

## Description

This PR address some feedback found in the iOS layer of the previous work done in PR691. This PR also points to `ci-interim` instead of `main` while CI is being sorted out.

## Breakdown

- Address iOS map leak.
- Implement some e2e tests.
- Fix JS linting issues.

## Validation

- Manually ran e2e tests on iOS and Android devices.

## Additional Notes

N/A

### DIFF
--- a/android/src/main/java/com/twiliovoicereactnative/ExpoModule.kt
+++ b/android/src/main/java/com/twiliovoicereactnative/ExpoModule.kt
@@ -154,7 +154,52 @@ class ExpoModule : Module() {
       options: Map<String, Any>,
       promise: Promise ->
 
-      this@ExpoModule.moduleProxy.callInvite.accept(uuid, PromiseAdapter(promise))
+      val iceServers = HashSet<IceServer>()
+
+      val jsIceServers = options[CommonConstants.CallOptionsKeyIceServers] as ArrayList<*>?
+
+      jsIceServers?.forEach { jsIceServer ->
+        if (jsIceServer !is LinkedHashMap<*, *>) return@forEach
+
+        val serverUrl =
+          jsIceServer[CommonConstants.IceServerKeyServerUrl] as String?
+        val username =
+          jsIceServer[CommonConstants.IceServerKeyUsername] as String?
+        val password =
+          jsIceServer[CommonConstants.IceServerKeyPassword] as String?
+
+        if (serverUrl != null && username != null && password != null) {
+          iceServers.add(IceServer(serverUrl, username, password))
+        } else if (serverUrl != null) {
+          iceServers.add(IceServer(serverUrl))
+        }
+      }
+
+      val jsIceTransportPolicy = options[CommonConstants.CallOptionsKeyIceTransportPolicy] as String?
+
+      val iceTransportPolicy = when (jsIceTransportPolicy) {
+        CommonConstants.IceTransportPolicyValueAll -> IceTransportPolicy.ALL
+        CommonConstants.IceTransportPolicyValueRelay -> IceTransportPolicy.RELAY
+        else -> null
+      }
+
+      var iceOptions: IceOptions? = null
+
+      if (iceServers.isNotEmpty() || iceTransportPolicy != null) {
+        val iceOptionsBuilder = IceOptions.Builder()
+
+        if (iceServers.isNotEmpty()) {
+          iceOptionsBuilder.iceServers(iceServers)
+        }
+
+        iceTransportPolicy?.let {
+          iceOptionsBuilder.iceTransportPolicy(it)
+        }
+
+        iceOptions = iceOptionsBuilder.build()
+      }
+
+      this@ExpoModule.moduleProxy.callInvite.accept(uuid, iceOptions, PromiseAdapter(promise))
     }
 
     AsyncFunction("callInvite_reject") {
--- a/android/src/main/java/com/twiliovoicereactnative/ExpoModule.kt
+++ b/android/src/main/java/com/twiliovoicereactnative/ExpoModule.kt
@@ -154,50 +154,9 @@ class ExpoModule : Module() {
       options: Map<String, Any>,
       promise: Promise ->
 
-      val iceServers = HashSet<IceServer>()
-
       val jsIceServers = options[CommonConstants.CallOptionsKeyIceServers] as ArrayList<*>?
-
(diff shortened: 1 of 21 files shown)


ios/TwilioVoiceReactNative+CallKit.m
```diff
@@ -147,22 +160,51 @@ - (void)makeCallWithAccessToken:(NSString *)accessToken
...
+            // `performVoiceCallWithUUID`, which never runs when the transaction
+            // fails, so clear them here.
+            //
+            // NOTE: `iceServersMap` and `iceTransportPolicyMap` are
+            // unsynchronized `NSMutableDictionary`s accessed from more than one
+            // queue -- written from the React Native bridge queue (here, and in
+            // `callInvite_accept`), read and removed on the main queue (the
+            // `perform*` methods, where read-then-remove is also not atomic),
+            // and removed here on whichever queue CallKit invokes this block on.
+            // Concurrent mutation of an `NSMutableDictionary` is unsafe.
```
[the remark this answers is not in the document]
mhuynh5757: Noted in the ticket referenced in the comment. VBLOCKS-7040

ios/TwilioVoiceReactNative+CallKit.m
```diff
@@ -120,6 +120,19 @@ - (void)answerCallInvite:(NSUUID *)uuid
 
     [self.callKitCallController requestTransaction:transaction completion:^(NSError *error) {
         if (error) {
+            // NOTE: this failure path never invokes `self.callKitCompletionCallback`,
+            // which has two consequences:
+            //
+            //   1. The completion block in `callInvite_accept` never runs, so the
+            //      promise returned by `CallInvite.accept()` never settles.
```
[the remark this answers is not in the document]
mhuynh5757: Known gap and filed under VBLOCKS-7039

test/appium-harness/src/test-suites/ice-test.ts
```diff
@@ -0,0 +1,441 @@
...
+      ): VariantResult => ({
+        variant: variantName,
+        outcome: actual === expectation ? 'passed' : 'failed',
+        expected: expectation,
+        actual,
+        note,
+      });
+
+      const connectResult = await safelySettlePromise(
+        voice.connect(token, options),
```
charliesantos: Every variant here calls `voice.connect` (PR691's outgoing path) - no variant accepts a call invite, so the `callInvite.accept` ICE plumbing added in this PR has zero end-to-end coverage. Add an incoming-call variant that calls `callInvite.accept({ iceServers: [BOGUS_ICE_SERVER], iceTransportPolicy: IceTransportPolicy.Relay })` and expects `Call.Event.ConnectFailure`, exercising the new iOS `performAnswerVoiceCallWithUUID` and Android `VoiceService.acceptCall` ICE paths.

android/src/main/java/com/twiliovoicereactnative/ExpoModule.kt
```diff
@@ -154,7 +154,11 @@ class ExpoModule : Module() {
       options: Map<String, Any>,
       promise: Promise ->
 
-      this@ExpoModule.moduleProxy.callInvite.accept(uuid, PromiseAdapter(promise))
+      val jsIceServers = options[CommonConstants.CallOptionsKeyIceServers] as ArrayList<*>?
```
charliesantos: `as ArrayList<*>?` and `as String?` are checked casts - a non-null value of an unexpected type (e.g. a `ReadableNativeArray` wrapper from the bridge) throws `ClassCastException` and crashes the JS engine thread. Change these to `as? ArrayList<*>` / `as? String` here and in `buildIceOptions` (lines 403-405), matching the `hasKey`/`Optional` guards the Java module uses; a null result is the safe no-option fallback.

test/appium-harness/src/test-suites/ice-test.ts
```diff
@@ -0,0 +1,441 @@
...
+
+/**
+ * Fill this in with a real TURN server to run the `valid-*` variants. Twilio's
+ * Network Traversal Service will hand these out. Left unfilled, those variants
+ * are skipped.
+ *
+ * Consider adding a mechanism to read this value out of env vars in a real CI
+ * pipeline.
+ */
+const VALID_ICE_SERVER = {
```
charliesantos, line 59: `VALID_ICE_SERVER` is hardcoded to `'TODO'`, and the skip guard (`serverUrl !== 'TODO'`) silently skips the `valid-servers-relay-policy` and `valid-and-bogus-servers-relay-policy` variants in every run - the only cases that verify establishment against a real TURN server. Read these from env vars (`process.env.ICE_SERVER_URL`, etc.) with `'TODO'` as fallback and document them in the harness README, otherwise those variants are dead code.

src/__tests__/CallInvite.test.ts
```diff
@@ -261,8 +264,7 @@ describe('CallInvite class', () => {
   describe.each([
```
charliesantos, line 264: The `.accept()` matrix dropped the `[{ foo: 'bar' }, { foo: 'bar' }]` row that verified unknown options forward to the native layer unchanged. TypeScript strips unknown keys at compile time only - at runtime they still pass through, and that contract is now untested. Worth adding a row cast `as any` that asserts the native module still receives the unknown key.

charliesantos (changes requested):
The iOS/Android accept-path ICE plumbing looks correct, but the e2e suite validates `voice.connect` (PR691's outgoing path), not `callInvite.accept` - the new feature has no end-to-end coverage. Separately, the Expo Kotlin module uses unchecked casts across its connect/accept/preflight paths where the Java module guards with `hasKey`/`Optional`; worth aligning the whole module on the safe pattern. (4 items from the local review were omitted as already raised by Copilot or already fixed.)

test/appium-harness/src/test-suites/ice-test.ts
mhuynh5757: `process.env` won't quite work for this gap since this is the client side JS that will be bundled into the release apk/ipa for the device. I have filed a ticket for investigating how to do this for CI, I left it this way for now for local testing.

bobiechen-twilio (approved):

charliesantos (approved):

PR #717, opened 2026-09-01
[VBLOCKS-7114] Increase e2e coverage

mhuynh5757:
## Submission Checklist

## Description

Addresses https://twilio-engineering.atlassian.net/browse/VBLOCKS-7114

## Breakdown

This PR adds increased e2e test coverage for the RN SDK in the Expo Appium Test Harness.

## Validation

- Ran all e2e test suites locally.

## Additional Notes

N/A

### DIFF
--- a/test/appium-harness/.gitignore
+++ b/test/appium-harness/.gitignore
@@ -53,5 +53,5 @@ app-example
 ###
 # Tokens
 ###
-/src/utilities/token/e2e-*-token*.ts
-/src/utilities/token/e2e-tests-ice-server*.ts
+/src/utilities/token/e2e-*-token*
+/src/utilities/token/e2e-tests-ice-server*
--- a/test/appium-harness/.gitignore
+++ b/test/appium-harness/.gitignore
@@ -48,3 +48,10 @@ app-example
 ##
 /secrets.json
 /google-services.json
+
+
+###
+# Tokens
+###
+/src/utilities/token/e2e-*-token*.ts
+/src/utilities/token/e2e-tests-ice-server*.ts
--- a/test/appium-harness/package.json
+++ b/test/appium-harness/package.json
@@ -4,7 +4,9 @@
   "version": "1.0.0",
   "scripts": {
     "start": "expo start",
-    "prebuild": "expo prebuild"
+    "prebuild": "expo prebuild",
+    "check:type": "tsc --noEmit",
+    "check": "yarn run check:type"
   },
   "dependencies": {
     "@react-navigation/bottom-tabs": "^7.15.5",
--- a/test/appium-harness/src/app/index.tsx
+++ b/test/appium-harness/src/app/index.tsx
@@ -93,9 +93,14 @@ export const Application = () => {
       'incoming-ice-test': incomingIceTest.perform,
     };
 
-    const perform = suites[testSuiteId as TEST_SUITE_ID] as
-      | (() => Promise<void>)
-      | undefined;
+    // `hasOwnProperty` rather than a plain lookup so that a suite id naming an
+    // inherited `Object.prototype` member - `toString`, `valueOf`,
+    // `constructor` - does not resolve to that member and skip the guard below.
+    const perform = Object.prototype.hasOwnProperty.call(suites, testSuiteId)
+      ? (suites[testSuiteId as TEST_SUITE_ID] as
+          | (() => Promise<void>)
+          | undefined)
+      : undefined;
 
     if (typeof perform === 'undefined') {
       logging.log.error(JSON.stringify({
--- a/test/appium-harness/src/app/index.tsx
+++ b/test/appium-harness/src/app/index.tsx
@@ -1,12 +1,22 @@
 import * as React from 'react';
-import { Button, Text, TextInput } from 'react-native';
+import { Button, Platform, Text, TextInput } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import { useVoice } from '../hooks/useVoice';
 import { useLogging } from '../hooks/useLogging';
+import { useCallControlsTest } from '../test-suites/call-controls';
+import { useCallMessageTest } from '../test-suites/call-message';
+import { useConnectOptionsTest } from '../test-suites/connect-options';
+import { useErrorsTest } from '../test-suites/errors';
+import { useIncomingCallManualTest } from '../test-suites/incoming-call-manual';
 import { useOutgoingCallTest } from '../test-suites/outgoing-call';
 import { useIceTest } from '../test-suites/ice-test';
 import { useIncomingIceTest } from '../test-suites/incoming-ice-test';
+import { usePreflightTest } from '../test-suites/preflight';
+import { useQualityWarningsTest } from '../test-suites/quality-warnings';
+import { useRegistrationTest } from '../test-suites/registration';
+import { useVoiceApiTest } from '../test-suites/voice-api';
 import { TestStatus } from '../test-suites';
+import { getPreflightTestToken } from '../utilities/token/get-token';
 
 /**
  * NOTE: VBLOCKS-6582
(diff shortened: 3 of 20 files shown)


test/appium-harness/.gitignore
```diff
@@ -48,3 +48,10 @@ app-example
 ##
 /secrets.json
 /google-services.json
+
+
+###
+# Tokens
+###
+/src/utilities/token/e2e-*-token*.ts
```
kpchoy: These are anchored to `.ts`, but `get-token.ts` requires without an extension and Metro resolves `js/jsx/ts/tsx/json`, each crossed with the platform suffixes. I checked with `git check-ignore`: every `.ts` form is covered including `.ios.ts`/`.android.ts`, but these are not:

```
e2e-tests-token.js
e2e-preflightTest-token.ios.js
e2e-tests-ice-server.json
```

Each of those loads exactly as well as the documented `.ts`. Given the doc asks testers to hand-create these before every session, and most token-minting snippets are Node/JS, a `.js` drop-in feels like a realistic way to land a live JWT in the repo.

Dropping the extension anchor covers it:

```
/src/utilities/token/e2e-*-token*
/src/utilities/token/e2e-tests-ice-server*
```

Or, fail closed:

```
/src/utilities/token/*
!/src/utilities/token/get-token.ts
```

test/appium-harness/src/app/index.tsx
```diff
@@ -16,46 +26,116 @@ import { TestStatus } from '../test-suites';
...
   const outgoingCallTest =
     useOutgoingCallTest(token, voice, logging, setTestStatus);
   const iceTest =
     useIceTest(token, voice, logging, setTestStatus);
   const incomingIceTest =
     useIncomingIceTest(token, voice, logging, setTestStatus);
+  const incomingCallManualTest =
+    useIncomingCallManualTest(token, voice, logging, setTestStatus);
 
   const performTest = React.useCallback(() => {
```
kpchoy, line 80: Two things on `performTest`, both cheap to close.

There's no in-flight guard, so a double tap on `button_startTestSuite` (or an Appium retry) starts a second `perform()` over the shared `Voice` from `useVoice`. Both runs then mutate device-wide state and race on the same `setTestStatus`, so the final status reflects whichever finishes last. That mattered less with three short suites; `preflight-test` now runs past two minutes and `incoming-call-test-manual` runs nine variants at 120s each.

```ts
if (testStatus === 'in-progress') {
  logging.log.warn(JSON.stringify({
    message: 'a test suite is already running',
  }));
  return;
}
```

(with `testStatus` added to the deps). Gating on `testStatus` rather than the returned promise is the right signal, since `outgoing-call-test` resolves its `perform` from an event listener before the suite is actually done.

Second, `return perform()` floats with no rejection handler at `onPress`. Every suite guards this internally and several say so in comments, which works today, but it's an invariant re-enforced by hand in twelve files instead of once here:

```ts
return perform().catch((error) => {
  logging.log.error(JSON.stringify({
    message: 'test suite rejected',
    testSuiteId,
    error: String(error),
  }));
  setTestStatus('failure');
});
```

test/appium-harness/src/test-suites/call-controls.ts
```diff
@@ -0,0 +1,366 @@
...
+      }
+
+      const to = call.getTo();
+      if (typeof to !== 'undefined') {
+        expect(to, 'call.getTo()').toBeTypeOf('string');
+      }
+    },
+  },
+  {
+    name: 'mute',
```
kpchoy, line 110: The header at line 27 says "any step that mutates call state restores it." `mute` and `hold` don't. If `expect(call.isMuted()).toBe(true)` throws, which is exactly the regression this step exists to catch, the call stays muted and `send-digits`, `get-stats` and `post-feedback` all run against it afterwards. A held call carries no media, so `get-stats` then fails for a reason that has nothing to do with `getStats`, which is the masking the header says can't happen.

`select-audio-device` in `voice-api.ts` already has the right shape:

```ts
run: async (call) => {
  try {
    expect(await call.mute(true), 'call.mute(true)').toBe(true);
    expect(call.isMuted(), 'call.isMuted() while muted').toBe(true);

    await delay(CONTROL_SETTLE_DELAY_MS);

    expect(await call.mute(false), 'call.mute(false)').toBe(false);
    expect(call.isMuted(), 'call.isMuted() while unmuted').toBe(false);
  } finally {
    // Swallowed: this is cleanup, and letting it throw would replace the
    // assertion failure that brought us here.
    await safelySettlePromise(call.mute(false));
  }
},
```

Same for `hold` with `call.hold(false)`.

test/appium-harness/src/test-suites/outgoing-call.ts
```diff
@@ -44,16 +45,36 @@ export const useOutgoingCallTest: UseTestSuite = (
       call.on(eventName, bindOnCallEvent(eventName));
     });
 
-    call.on(Call.Event.ConnectFailure, () => {
-      setTestStatus('failure');
+    let settled = false;
```
kpchoy, line 52: `settle` is only reachable from the `ConnectFailure` and `Disconnected` listeners, and nothing in the suite calls `call.disconnect()`, so it depends entirely on the far end ending the call. If neither event arrives (TwiML that never terminates, media path drops after connect), `settle` never runs: no summary, `setTestStatus` never called again, `text_testSuiteStatus` reads `in-progress`, and the call is left live on the device for whatever suite runs next.

Worth closing because this is the one suite the orchestrator actually drives today, and it's the only suite left without a bounded terminal wait now that the others all have one.

```ts
const timeoutId = setTimeout(async () => {
  await safelySettlePromise(call.disconnect());
  settle('failed', `no terminal call event within ${CALL_TIMEOUT_MS}ms`);
}, CALL_TIMEOUT_MS);
```

with `clearTimeout(timeoutId)` inside `settle`, next to the existing `settled` guard.

test/appium-harness/src/test-suites/ice-test.ts
```diff
@@ -49,20 +50,13 @@ const BOGUS_ICE_SERVER = {
...
  */
-const VALID_ICE_SERVER = {
-  serverUrl: 'TODO',
-  username: 'TODO',
-  password: 'TODO',
-};
+const VALID_ICE_SERVER = getIceServer();
 
-const HAS_VALID_ICE_SERVER = VALID_ICE_SERVER.serverUrl !== 'TODO';
+const HAS_VALID_ICE_SERVER =
```
kpchoy, line 58: Following on from this flag: when it's false the `valid-*` variants are skipped, but `skipped` never reaches the pass/fail decision. Line 436 computes it only to log it, same in `incoming-ice-test.ts:550`, and `summarizeResults` hands back a `skipped` count that the ten `runSteps` suites also discard.

The practical effect: `e2e-tests-ice-server.ts` is gitignored, so on a fresh clone every `valid-*` variant skips and the suite reports `success`. Those variants are the only ones proving `iceServers` credentials are actually applied on the connect and accept paths, and the orchestrator only reads `text_testSuiteStatus`, so "ran with relay coverage" and "ran with relay coverage silently dropped" are the same green.

Simplest version: gate on an explicit opt-out so an unintended credential gap fails rather than skips, and let the deliberate case stay quiet. Alternatively compare skipped names against an expected-skips set for the platform and fail on anything outside it.

test/appium-harness/src/utilities/run-steps.ts
```diff
@@ -0,0 +1,171 @@
...
+  log.info(JSON.stringify({
+    running: step.name,
+    description: step.description,
+  }));
+
+  // `Promise.resolve().then(...)` so that a `run` that throws synchronously
+  // becomes a failed step instead of rejecting `runStep`. A rejected `runStep`
+  // rejects `runSteps` and then the suite's `perform`, and nothing awaits
+  // `perform`, so the status would be left at `in-progress` with no summary.
+  const result = await safelySettlePromise(
```
kpchoy, line 99: The `Promise.resolve().then(...)` wrapper handles a step that *throws*, and the comment above it explains why. A step that never *settles* isn't covered, and the outcome is the one the comment says this exists to prevent: `runSteps` never returns, `summarizeResults` never runs, and the suite sits at `in-progress` with no summary at all, which is a worse signal than a plain failure.

Several step bodies await native promises with no deadline of their own: `voice.register`, `call.mute`/`hold`/`sendDigits`/`getStats`, `audioDevice.select`, `call.sendMessage`. `registration.ts:192` already races its teardown for this reason, so the pattern is here, just applied to one step rather than the runner.

```ts
const timeoutMs = step.timeoutMs ?? DEFAULT_STEP_TIMEOUT_MS;

const result = await Promise.race([
  safelySettlePromise(Promise.resolve().then(() => step.run(context, log))),
  delay(timeoutMs).then(() => 'timeout' as const),
]);

if (result === 'timeout') {
  return {
    step: step.name,
    outcome: 'failed',
    note: `did not settle within ${timeoutMs}ms`,
  };
}
```

with an optional `timeoutMs` on `Step<C>` for the slow ones. Worth noting in the docstring that the body keeps running after the race is lost, so steps shouldn't leave device-wide state behind.

test/appium-harness/src/utilities/token/get-token.ts
```diff
@@ -0,0 +1,81 @@
...
+  username: string;
+  password: string;
+};
+
+function readToken(tokenModule: unknown): string {
+  const token = (tokenModule as { token?: unknown } | undefined)?.token;
+  return typeof token === 'string' ? token : '';
+}
+
+export function getToken(): string {
```
kpchoy, line 29: `getToken` has no callers. `index.tsx:47` uses `getPreflightTestToken`, with a comment saying that's deliberate for every suite, and the two ICE suites take `getIceServer`. So the `console.warn` here can never fire.

The reason I'd rather not leave it: the file header uses `e2e-tests-token.android.ts`/`.ios.ts` as its worked example, and the new gitignore rule covers that name, so it reads as sanctioned. `MANUAL_E2E_TESTING.md` doesn't list it. Someone setting up locally will create the module, see no effect, and have no way to tell which source is stale.

Suggest deleting it and retitling the header example around `e2e-preflightTest-token`, so the code, the ignore rules and the doc table name the same set. If a second non-preflight token is genuinely planned, happy to see it land with its first caller instead.

test/appium-harness/src/utilities/token/get-token.ts
```diff
@@ -0,0 +1,81 @@
...
+    console.warn('No bundled e2e test token; enter one in the app.');
+    return '';
+  }
+}
+
+export function getPreflightTestToken(): string {
+  try {
+    // @ts-ignore - the module is a gitignored local secret and may not exist
+    return readToken(require('./e2e-preflightTest-token'));
+  } catch (error) {
```
kpchoy: This catch (and the ones at 33 and 77) swallows every throwable, and the `error` binding is unused, so the warning asserts the module was absent no matter what actually happened. A module that exists but throws (a syntax slip after a hand-edit, a wrong-shaped export) is indistinguishable from a missing one, and the only diagnostic says the opposite of the truth.

Because `index.tsx:47` now seeds all twelve suites from `getPreflightTestToken()`, a broken token module quietly gives every suite `''`, and the first symptom is a connect error in a suite that has nothing to do with tokens. `voice-api-test` and `errors-test` will even report `success` on an empty token, since neither needs a valid one.

Including the real error would cover most of it:

```ts
} catch (error) {
  console.warn(`No bundled preflight test token: ${String(error)}`);
  return '';
}
```

And a check where the cause is still known, either in `Application` or at the top of each token-dependent `perform`:

```ts
if (token.length === 0) {
  log.error(JSON.stringify({
    message: 'no access token; see src/utilities/token/get-token.ts',
  }));
  setTestStatus('failure');
  return;
}
```

test/appium-harness/src/utilities/expect.ts
```diff
@@ -0,0 +1,336 @@
...
+   */
+  const fail: (expectation: string) => never = (expectation) => {
+    throw new AssertionError(`expected ${subject} ${expectation}`);
+  };
+
+  return {
+    toBe: (expected) =>
+      check(Object.is(received, expected), `to be ${format(expected)}`),
+
+    toStrictEqual: (expected) =>
```
kpchoy, line 231: Five of the thirteen matchers have no call site anywhere under `src/`: `toStrictEqual`, `toBeTruthy`, `toBeFalsy`, `toBeUndefined`, `toMatch`. (For contrast, `toBe` has 62, `toBeTypeOf` 31, `toBeInstanceOf` 23.)

The cost isn't just those five declarations. `deepStrictEqual` at line 100 exists for `toStrictEqual`, and its only other consumer is `toContain`'s array branch, where all seven call sites pass a string or primitive, so nothing reaches past the `Object.is` leaf. The `RegExp` branch of `format` serves only `toMatch`, and the `bigint`/`symbol` branches have no reachable caller either. That's close to 100 of 336 lines.

Since the header argues for keeping the surface small and nothing external depends on this yet, suggest trimming to what the suites use and adding matchers back on demand. `toContain`'s array branch becomes `received.includes(expected)`.

test/appium-harness/package.json
```diff
@@ -4,7 +4,9 @@
   "version": "1.0.0",
   "scripts": {
     "start": "expo start",
-    "prebuild": "expo prebuild"
+    "prebuild": "expo prebuild",
+    "check:type": "tsc --noEmit",
+    "check": "yarn run check:type"
```
kpchoy, line 9: Nothing invokes these. CI runs `yarn --cwd test/appium-harness install --immutable` and then only the root `yarn run check`, which chains its own targets and never delegates to a sub-project. Neither `test/app` nor `test/expo` defines a `check`, so there's no existing convention being followed either.

They aren't redundant though, which is why I'd wire them rather than drop them: root `tsconfig.json` has no `include`/`exclude`, so root `tsc --noEmit` does type-check these files, but under the root config rather than the harness's own `expo/tsconfig.base`. The harness's real build config is the one nothing checks.

```yaml
- run: yarn --cwd test/appium-harness check
```

after the existing install step. Unrelated but adjacent: `test/appium-harness/` is in the root `eslintIgnore`, so the `eslint-disable jest/valid-expect` comment at the top of `call-controls.ts` is currently inert.

kpchoy (changes requested):
Really nice piece of work, and a big lift. The teardown discipline, the consistent `safelySettlePromise` use, and comments that explain *why* rather than *what* made this genuinely pleasant to read through.

charliesantos (dismissed):

test/appium-harness/src/test-suites/outgoing-call.ts
mhuynh5757: Since the default token is the PreflightTest one, the other end does indeed end the call. But if something is wrong or the token points to the wrong Twiml App, this is a valid concern. 👍

kpchoy (dismissed):

test/appium-harness/src/app/index.tsx
mhuynh5757: Added a comment to revisit this one later.

test/appium-harness/src/test-suites/call-controls.ts
mhuynh5757: Added comment for this one as well.

test/appium-harness/src/test-suites/outgoing-call.ts
mhuynh5757: Added a comment for this one as well.

test/appium-harness/src/test-suites/ice-test.ts
mhuynh5757: Added a comment.

test/appium-harness/src/utilities/run-steps.ts
mhuynh5757: Added a comment.

test/appium-harness/src/utilities/expect.ts
mhuynh5757: Added a comment.

test/appium-harness/package.json
mhuynh5757: I think this is actually OK to leave as-is. We don't want failures inside the test projects to fail the build for the actual library code.

kpchoy (approved):

PR #718, opened 2026-09-09
[VBLOCKS-7113] Fix ga api gaps

mhuynh5757:
## Submission Checklist

## Description

Closes API gaps found while auditing the public surface against the native layers. Most changes are breaking, and each one is documented in `CHANGELOG.md`.

## Breaking changes

- `Call.getStats` resolves with an array of `RTCStats.StatsReport`, matching what the native layer has always returned.
- Fixed three misspelled `RTCStats` fields: `requestsReceieved`, `responsesRecieved`, and `bytesRecieved`. These were always `undefined` at runtime.
- `PreflightTest.getEndTime` resolves with `undefined` when the test has not ended. `PreflightTest.getLatestSample` and `PreflightTest.getReport` are unchanged, and their documentation now states what they resolve with before a value exists.
- Removed `CallInvite.isValid`, an unused alpha method.
- `IncomingCallMessage.getContent` returns `string` instead of `any`.
- `Voice.showAvRoutePickerView` rejects with an `UnsupportedPlatformError` on Android instead of silently resolving.
- `Voice.getDeviceToken` rejects with an `InvalidStateError` on iOS when no Device token is available instead of resolving with an empty string.
- Added `Call.QualityWarning.ConstantAudioOutputLevel`, renamed `Call.QualityWarning.HighPacketLoss` to `HighPacketsLostFraction`, and stopped iOS from reporting unrecognized warnings as `"undefined"`.
- Added `AudioDevice.Type.WiredHeadset`. A wired headset previously reported `Earpiece` on Android and `Unknown` on iOS.

## Fixes

- `CallInvite.reject` now rejects when the native layer reports an error. Previously the returned `Promise` resolved regardless of the outcome.

## Tests

Unit test coverage for each of the above. The Appium harness gains suites for audio devices, RTC stats, call invite rejection, preflight early state, and ICE, along with attended and unattended suite runners.


kpchoy:
## Review Summary

A careful, well-evidenced PR: the `RTCStats` spelling fixes, the `getStats` array cardinality, the `Call.QualityWarning` rename, and the `CallInvite.isValid` removal were all verified correct against the pinned native SDKs (`voice-android:6.10.3`, `TwilioVoice 6.13.6`), the generated-constants rule was respected, and `api/voice-react-native-sdk.api.md` regenerates with zero drift. 546 unit tests, lint, and `check:api` all pass.

The findings cluster in two places. First, a load-bearing factual claim — "the native layer reports 'no sample yet' as an all-zero sample, not as nil, so this path is unreachable today" — is contradicted by the vendored `TVOPreflight.h`, which declares both `latestSample` and `preflightReport` `nullable`. That claim has been propagated into public TSDoc, `src/type/NativeModule.ts`, the CHANGELOG, and a test fixture, so the tests confirm the assumption rather than test it. Second, several gaps are documented with ticket references rather than fixed — including `AudioDevice.select()`, whose one-line fix is the same one this PR applies to `CallInvite.reject` two files over, and which makes the newly added `AudioDevice.Type.WiredHeadset` undetectably broken on iOS.

16 findings: 1 critical, 15 high.

<details><summary>🔴 Critical Findings (1)</summary>

#### "Unreachable today" contradicts the vendored nullable headers — `ios/TwilioVoiceReactNative+PreflightTest.m:72`
_[Found by 4 agents: comment-analyzer, silent-failure-hunter, docs-validator, language-pattern-reviewer]_

`preflightTest_getLatestSample` (line 72) and `preflightTest_getReport` (line 94) each claim the nil path is "unreachable today" because the native layer reports an all-zero sample. The same file, ~100 lines lower, states the opposite as justification for the nil guards this PR adds: "The sample is nil until the preflight test generates its first one" (line 174) and "The report is nil until one is available" (line 213). Both cannot be true.

The pinned TwilioVoice 6.13.6 headers settle it:

```objc
// TVOPreflight.h
@property (nonatomic, strong, readonly, nullable) TVOPreflightReport *preflightReport;    // line 55
@property (nonatomic, strong, readonly, nullable) TVOPreflightStatsSample *latestSample;  // line 60
```

So on iOS the routine pre-first-sample state reaches the new guard and rejects with `InvalidStateError` carrying the message "could not serialize" — a normal state misreported as a serialization defect. The claim has been propagated into five places: `src/PreflightTest.ts` public TSDoc for `getLatestSample`/`getReport`, `src/type/NativeModule.ts:140` ("this always resolves with a JSON string"), `android/.../PreflightTestModuleProxy.java:80,98`, `CHANGELOG.md:38` ("There is no separate absent state to detect"), and `src/__mock-data__/PreflightTest.ts:270` ("Both platforms report **this exact** all-zero-valued sample/report"), where it is hardened into a fixture. The new tests mock that payload, so they pass regardless of what the native SDKs do.

Consumers are told explicitly not to handle a rejection that iOS will produce.

**Suggestion:** Verify on-device what `latestSample`/`preflightReport` return before the first sample. Then either (a) have iOS synthesize the all-zero payload so the documented cross-platform contract actually holds, or (b) keep the reject and correct the TSDoc, `NativeModule.ts`, and CHANGELOG to document `InvalidStateError` as a real outcome — and give it a message that names the state ("has not generated a sample yet") rather than blaming serialization. Either way, drop the "unreachable today" comments and remove "this exact" from the fixture docblock.

</details>

<details><summary>🟠 High Findings (15)</summary>

#### `AudioDevice.select()` still swallows every native failure — `src/AudioDevice.tsx:103`
_[Found by 5 agents: bug-hunter, code-reviewer, silent-failure-hunter, type-design-reviewer, comment-analyzer]_

`select()` awaits the raw `NativePromise` envelope instead of unwrapping it with `settleNativePromise`, so every failure resolves successfully — while the TSDoc three lines above still promises "Rejects if the audio device cannot be selected." Both platforms have live failure paths: `VoiceModuleProxy.selectAudioDevice:119-124` rejects on an unknown uuid, and `ios/TwilioVoiceReactNative.m:713-723` rejects when the port lookup fails. The Android path is routinely reachable — `AudioSwitchManager.start()` mints a fresh `UUID.randomUUID()` per device on every route change, so any handle held across a headset plug/unplug has a dead uuid.

This PR applies the exact one-line fix to `CallInvite.reject` in the same diff, and the added comment spells the fix out. Leaving it deferred also makes the new `AudioDevice.Type.WiredHeadset` undetectably broken on iOS: the new `selectAudioDevice:` wired-headset branch is documented by its own comment (VBLOCKS-7140) as unable to succeed, and `select()` reports that permanent failure as success.

**Suggestion:** `await settleNativePromise(NativeModule.voice_selectAudioDevice(this.uuid));`, add the rejection test (the new `CallInvite.reject` test is the template), and drop the comment — the existing TSDoc then becomes true. If the deferral is deliberate, the TSDoc must be corrected in this PR so the published contract is not false for the life of VBLOCKS-7139.

#### Dropped terminal `Completed` event leaves consumers hung — `ios/TwilioVoiceReactNative+PreflightTest.m:354`
_[Found by 3 agents: silent-failure-hunter, bug-hunter, code-reviewer]_

`preflight:didCompleteWithReport:` now returns early with only an `NSLog` when serialization yields nil. `Completed` is terminal — the SDK will not call `didFailWithError:` afterwards — so the JS `PreflightTest` emits neither `Completed` nor `Failed`, and anything awaiting a terminal event waits forever with nothing on the JS side. The previous behavior (`NSInvalidArgumentException`) was also wrong but loud; this converts a crash into a permanent silent hang, which inverts the PR's stated goal.

The trigger is reachable: `dictionaryReport` carries `double` members (`mos`, `packetsLostFraction`, and the `TVOPreflightAggregateStats` average/min/max triples), and NaN or infinity — what a degenerate run with zero samples produces — makes serialization fail. The same early return for `didReceiveStatsSample:` is acceptable, since samples are periodic and non-terminal.

**Suggestion:** Emit `kTwilioVoiceReactNativePreflightTestEventTypeValueFailed` with the error payload that event already carries, so the JS layer reaches a terminal state, and propagate the underlying `NSError` into the message.

#### `NSJSONSerialization` raises; it does not return an `NSError` here — `ios/TwilioVoiceReactNative+PreflightTest.m:226`

Both hardened serializers treat `+dataWithJSONObject:options:error:` as error-returning. For an object that is not a valid JSON object it **raises `NSInvalidArgumentException`** instead — so the new `if (jsonData == nil)` guards cannot be reached for the failure mode that matters (NaN/infinity in the SDK-controlled doubles above). The process crashes before either check runs, which means the new nil-guards give false confidence: the comments say a serialization failure now surfaces as a rejected promise or a dropped event, but a real one still crashes the app.

**Suggestion:** Guard with `+[NSJSONSerialization isValidJSONObject:]` (and/or `@try`) before serializing in both `preflightStatsSampleToJsonString` and `preflightReportToJsonString`, so the real failure routes into the `nil` return the callers were just taught to handle.

#### CHANGELOG claims `CallInvite.reject` rejects; iOS never does — `CHANGELOG.md:104`
_[Found by 3 agents: docs-validator, silent-failure-hunter, comment-analyzer]_

The unqualified Fixes entry says "`CallInvite.reject` now rejects when the native layer reports an error." The JS wrap is correct, but it only has effect where native produces a rejection envelope, and this PR's own diff adds the acknowledgement that iOS does not: `// VBLOCKS-7159 / This method resolves the JS promise unconditionally.` There is no lookup and no failure path there, so rejecting a stale, unknown, or already-ended invite still silently "succeeds" on iOS.

On Android the only rejecting path is `getCallRecord`'s pre-call guard (`InvalidArgumentError` for a missing/expired record) — not "the native layer reports an error" during rejection. Every other platform-conditional behavior in this CHANGELOG is correctly filed under "Platform Specific Fixes".

**Suggestion:** Either fix iOS `callInvite_reject` to look up the invite and reject (the `getCallInviteFromMap:` guard directly below is the shape), or move this entry under Platform Specific Fixes → Android and state that iOS still resolves unconditionally pending VBLOCKS-7159. Also add the newly possible rejection to `CallInvite.reject()`'s TSDoc, which still documents only the resolve.

#### New e2e suites can never report success — `test/appium-harness/src/test-suites/rtc-stats.ts:282`

Three new suites embed steps documented as deterministically failing against defects this PR does not fix, yet each computes `setTestStatus(failed === 0 ? 'success' : 'failure')`:

- `rtc-stats.ts:282` `ice-candidate-pair-state-is-a-known-value` (VBLOCKS-7160 — see next finding)
- `audio-device.ts:226` `select-wired-headset` (iOS, VBLOCKS-7140) and `:284` `select-a-removed-device-rejects` (both platforms, VBLOCKS-7139)
- `call-invite-reject.ts:254` `second-reject-rejects` (iOS, VBLOCKS-7159)

Because `summarizeSuiteResults` treats any non-`success` outcome as failed, both new aggregate runners — `unattended-all` and `attended-all` — are permanently red. With no green baseline, a genuine regression is indistinguishable from the known failures and a tester has to diff step names by hand.

**Suggestion:** Add an explicit expected-failure marker (`knownFailing: { ticket, platforms }`) that `runSteps` excludes from the `failed` count driving `setTestStatus`. That also makes a suite go red when a known-failing step starts *passing*, which is the signal the ticket is fixed.

#### `IceCandidatePairState` values never match native output — `src/type/RTCStats.ts:34`
_[Found by 2 agents: type-design-reviewer, test-analyzer]_

This PR fixes three `RTCStats` keys that never matched native. The identical gap sits one type away in the same file and is left in place: `IceCandidatePairState` declares `STATE_FAILED = 'STATE_FAILED'`, `STATE_SUCCEEDED = 'STATE_SUCCEEDED'`, etc., while both platforms emit the camelCase constants — `ios/Utilities/TwilioVoiceStatsReport.h:156` and `android/.../StatsListenerProxy.java:164` both return `constants.src`'s `stateFailed` / `stateSucceeded` / … So `stats.state === RTCStats.IceCandidatePairState.STATE_SUCCEEDED` is always false, on every platform. iOS additionally emits `stateUnknown`, which has no enum member at all.

These literals are published in `api/voice-react-native-sdk.api.md`, so fixing it later is another breaking change. This release is already taking the breaking hit on `RTCStats` field names and `Call.QualityWarning` values in the same CHANGELOG section — one more bullet now versus another major bump later. Tracked as VBLOCKS-7160, currently asserted only by the always-failing e2e step above.

**Suggestion:** Redefine the members from `Constants.State*` the way `Call.QualityWarning` now is, add `STATE_UNKNOWN`, mirror the `Call.test.ts` constant-agreement test pair, and add a CHANGELOG entry. Decide whether Android should also emit `StateUnknown` (its switch has no such arm).

#### `getEndTime` reports malformed native values as "has not ended" — `src/PreflightTest.ts:495`
_[Found by 2 agents: silent-failure-hunter, type-design-reviewer]_

`if (!Number.isFinite(endTime) || endTime === 0) return undefined;` folds two unrelated conditions together. `endTime === 0` is the documented "has not ended" signal and is correct. `!Number.isFinite(endTime)` means native sent something that is not a number — a native defect — now reported as the perfectly normal "still running" state, so a consumer polling for completion polls forever with nothing raised or logged. The method's TSDoc still says "Rejects if the native layer encountered an error."

This is the same conflation the PR removes elsewhere (iOS `getDeviceToken` no longer returns `""`, Android `showAvRoutePickerView` no longer no-ops), and the PR's own VBLOCKS-7141 comments argue malformed native payloads should surface via `constructInvalidValueError` — which this file uses nine times.

Separately, the `nativeEndTime === null || typeof nativeEndTime === 'undefined'` guard is unreachable against `preflightTest_getEndTime(): NativePromise<string>` in `src/type/NativeModule.ts:138`. Either the branch is dead or the native type should be `NativePromise<string | null>`; the new test passes `null` through an `any`-typed mock so it pins neither.

**Suggestion:** Split the branches — return `undefined` for `0`, throw `constructInvalidValueError` for a non-finite value — and reconcile the null guard with the declared native type.

#### `showAvRoutePickerView` resolves on iOS even when the picker is never shown — `src/Voice.tsx:773`

The PR removes the Android silent no-op (good) and rewrites the contract to "Resolves when the AV Route Picker View is shown." iOS does not honor it (`ios/TwilioVoiceReactNative.m:725-745`): when `rootViewController` is nil the picker is never presented and the promise still resolves — the same "failure indistinguishable from success" defect the CHANGELOG says this change fixes, still present on the only platform where the method is supported. Presentation is also `dispatch_async`'d, so resolution does not mean "shown" even on the success path. `windows[0]` additionally raises `NSRangeException` on an empty windows array, reachable in scene-based apps during backgrounding (pre-existing).

**Suggestion:** Reject with `InvalidStateError` when no window or root view controller is available instead of falling through to `resolvePromise`, and guard `windows[0]` with a count check. If resolving before presentation is intended, weaken the TSDoc to "has been requested".

#### All-zero sentinel is typed as a valid `Report`/`RTCSample` — `src/PreflightTest.ts:539`

`getEndTime` was widened to `Promise<number | undefined>` in this PR precisely because "no real end time" needed to be representable. Its two siblings on the same class keep `Promise<PreflightTest.Report>` / `Promise<RTCSample>` while the new fixtures spell out what they resolve with before the test is ready: `callSid: ''`, `edge: ''`, empty arrays, every `TimeMeasurement` zeroed. That value type-checks as a `Report` while satisfying none of a `Report`'s invariants, and TypeScript actively tells the consumer a check is unnecessary.

The CHANGELOG's answer — listen for `Event.Sample`/`Event.Completed` instead of inspecting the resolved value — puts a real invariant in prose on a GA API. It is also internally inconsistent: `getEndTime` translates the `"0"` sentinel because no real end time can be `0`, while `callSid: ''` and an all-zero `testTiming` are exactly as unambiguous and pass straight through.

**Suggestion:** Widen both to `| undefined` and detect the sentinel at the boundary (`if (!reportObj.callSid) return undefined;`, `timestamp === '0'` for the sample), or use a discriminated `{ status: 'ready' … } | { status: 'not-ready' }` that generalizes to `getEndTime` too. Note this interacts with the critical finding — settle the native behavior first.

#### Android catch logs nothing at all when `getMessage()` is null — `android/.../PreflightTestModuleProxy.java:50`

`logger.warning(error.getMessage())` loses the exception type and stack trace, and for the most likely `RuntimeException` here — an NPE from the native layer — `getMessage()` is null. `SDKLog.warning(String)` then NPEs on `message.getBytes()` inside its own `catch (Exception ignore) {}`, so nothing is logged at all and JS receives an `InvalidStateError` with a null message. The catch added to prevent a main-thread crash turns a diagnosable crash into an invisible failure. `SDKLog` already has a `warning(Exception, String)` overload that prints the stack trace, unused here.

(Two other agents examined this and confirmed it does not crash — `constructJSMap` filters null values — so this is a diagnosability regression, not a stability one.) Catching all of `RuntimeException` also relabels genuine bugs in this module's own lambdas, including the two `IllegalStateException`s the PR just added, as `InvalidStateError`.

**Suggestion:** Use `logger.warning(error, String.format("getPreflightTest(%s) failed", uuidStr))` and fall back to `error.toString()` when `getMessage()` is null.

#### Android quality-warning omission documented as iOS-only — `CHANGELOG.md:74`

The bullet is scoped "On iOS, … unrecognized warnings are omitted". Android changed the same way in this PR: `serializeCallQualityWarnings` no longer emits `warning.toString()` for every warning but maps through `callQualityWarningToString`, which returns null outside the six mapped cases, and those are dropped with a log. An Android consumer reading this section concludes only the value rename affects them. The explicit mapping also means Android-reported values are now the `constants.src` strings rather than whatever `toString()` produced, and the CHANGELOG never says which Android strings changed beyond the packet-loss rename.

**Suggestion:** Move the omission sentence out of the iOS-only bullet ("on both platforms… previously `"undefined"` on iOS and the raw native string on Android") and list any other Android-reported value that changed.

Worth noting for the native-SDK-upgrade checklist: both `default:` branches are unreachable at the pinned versions (all six enum members are mapped, verified against `voice-android:6.10.3` and TwilioVoice 6.13.6), so nothing is dropped today. A seventh warning in a future SDK bump would be silently omitted, and an all-unrecognized set would arrive as an empty array indistinguishable from "warnings cleared".

#### Two native crash fixes have no CHANGELOG entry — `android/.../PreflightTestModuleProxy.java:44`

Both are user-visible and belong in a release whose Fixes section already itemizes an Android NPE crash fix:

- Android: `getPreflightTest` now wraps the main-handler callback in try/catch — previously a throw from `getLatestSample()`/`getReport()` crashed the app; now the `Promise` rejects.
- iOS: `preflight:didReceiveStatsSample:` / `didCompleteWithReport:` previously raised `NSInvalidArgumentException`; now the event is dropped instead. That changes what an application can rely on receiving — which matters because the `getLatestSample`/`getReport` entry tells applications to rely on those events instead of the resolved value.

**Suggestion:** Add an Android entry ("`PreflightTest` methods no longer crash the application when the native layer throws; the `Promise` rejects with an `InvalidStateError` instead") and an iOS one describing the dropped-event behavior — ideally after resolving the terminal-event finding above.

#### Stale comment cites the Android `toString()` path this PR deleted — `src/__tests__/Call.test.ts:773`

The new docblock says to see "`Call.CallQualityWarning.toString()` on Android". This same PR replaced that with the explicit `callQualityWarningToString` switch, precisely because — as that function's own comment says — the platforms name the packet-loss warning differently below this layer. A maintainer debugging a warning mismatch would look for an enum-name-to-string relationship, find that `WARN_HIGH_PACKET_LOSS` does not stringify to `high-packets-lost-fraction`, and could "fix" the constant to match `toString()`, reintroducing the divergence this PR removed.

**Suggestion:** Point at `ReactNativeArgumentsSerializer.callQualityWarningToString` and note the mapping is explicit *because* the native enum's `toString()` does not match these values.

#### VBLOCKS-7141 comment states a false contrast — `src/PreflightTest.ts:662`

The `parseSample` comment asserts "Unlike `parseReport`, which throws on a partial payload, this surfaces nothing at all." `parseReport` reads `callSid`, `edge`, `iceCandidates`, `networkStats`, `selectedEdge` and `selectedIceCandidatePair` with bare property access and no validation — a payload missing any of them yields `undefined` members and throws nothing, exactly like `parseSample`. It throws only for `networkTiming` and `statsSamples`, which are dereferenced a level deeper. The two differ in which subset blows up, not in kind.

Both VBLOCKS-7141 comments then prescribe a `constructInvalidValueError` fix in enough detail to implement, in a file that already imports and uses that helper nine times, and the fix is not applied.

**Suggestion:** Correct or drop the false contrast. Then either apply the prescribed validation, or reduce both comments to a bare observation and let the ticket hold the fix plan — the plan is what rots, the observation is not.

#### Native-only changes have no automated coverage — `test/appium-harness/src/test-suites/unattended-all.ts:1`

The repo has no native unit tests (no `*Test.java`, no XCTest target), and `test/appium-orchestrator/test-suites/index.mjs:22` hard-codes `await outgoingCallTest(testElements)` as the only entry point — so of the twelve harness suites, exactly one runs in CI. This PR adds `unattended-all`, the natural fix, but does not wire it in. Changed native paths with no test that can fail anywhere in the pipeline: Android `callQualityWarningToString` (only the constant-audio warnings are provoked by the e2e TwiML), iOS `warningNameWithNumber`'s nil branch, iOS `audioPortTypeMapping` for `AVAudioSessionPortHeadphones`, the Android `AudioSwitchManager` WiredHeadset/Earpiece split, and the iOS `getDeviceToken` rejection (the `voice-api` step accepts either outcome, so it cannot distinguish the old contract from the new).

**Suggestion:** Wire `unattended-all` into the orchestrator (parameterize the suite id, defaulting to `unattended-all` in CI) — this depends on the known-failing-step fix above, or CI is permanently red. The Android warning mapping is pure and platform-independent, so an `android/src/test` JVM test is cheap insurance for what is now the only thing standing between a native enum and the public `Call.QualityWarning` values.

</details>

<details><summary>📋 Also noted, not filed as findings</summary>

- `test/appium-harness/docs/MANUAL_E2E_TESTING.md:115` — the "Suites that need an inbound call" section still says "Both inbound suites" and covers only `incoming-ice` and `incoming-call-manual`; the new `call-invite-reject` needs a procedure paragraph, and the new `audio-device` suite has no plug/unplug procedure anywhere. Line 128 also says `incoming-ice` waits 60s per variant while `attended-all.ts` cites a 15-minute ceiling.
- `src/__tests__/Call.test.ts:464` — the `getStats` array assertion replays the mock edited in the same commit, so it cannot fail for any value of the change. Unavoidable for a pass-through, but the rejection path (the one branch that can fail) is untested.
- `ios/TwilioVoiceReactNative.m:321-343` — the new wired-headset branch is byte-identical to the `unknown` branch below it and documented as non-functional; deleting it leaves behavior unchanged with one less thing to explain.
- `Call.QualityWarning` `toHaveLength(6)` proves JS agrees with JS only — `Constants` is generated from the same `constants.src`, so the native switches are invisible to it. Adding `expect(warningValues).toHaveLength(6)` would close the reverse direction.
- API Extractor emits pre-existing `ae-missing-release-tag` warnings for `PreflightTest`, the audio-codec types, and `IceTransportPolicy`/`IceServer` — the last of these is in a file this PR touches.
- Verified clean, for the record: all `parseReport` callers updated; `getStats` array correct on both platforms; no remaining nil-insertion path after `warningNameWithNumber` returns nil; both native quality-warning enums have all six members so the new switches compile; `Constants.X` as a string-enum initializer is legal and inlines correctly into the api report; `getContent(): string` is safe; `callInvite_isValid` removal is complete on both platforms; `docs/bare-rn-support-guide.md` needs no update since `src/common.ts` is untouched.

</details>

<details><summary>📋 Agents Dispatched</summary>

**Ran:** code-reviewer, bug-hunter, language-pattern-reviewer, silent-failure-hunter, type-design-reviewer, comment-analyzer, test-analyzer, docs-validator, test-runner

**Skipped:** security-reviewer (no auth, input handling, or crypto changes), security-scanner (no new attack surface), dependency-scanner (no manifest changes), performance-reviewer (no I/O or query changes), architecture-reviewer (no structural refactor), architecture-standards-reviewer (no new service boundaries), simplification-reviewer (change is corrective), api-standards-reviewer (no REST surface), coverage-scorer (covered by test-analyzer)

**Complexity tier:** complex

**test-runner results:** 546 tests / 17 suites pass, lint clean, `build:docs` + `git diff --exit-code src/error/generated.ts api/` shows zero API drift. The 15 `check:type` errors are all in the uninstalled `test/app`, `test/expo`, and `test/appium-harness` sub-projects — an environment artifact, not a PR defect.

</details>

---
<sub>Reviewed by [review-toolkit](https://github.com/twilio-internal/claude-marketplace/tree/main/plugins/review-toolkit)</sub>

ios/TwilioVoiceReactNative+PreflightTest.m
```diff
@@ -69,6 +69,16 @@ @implementation TwilioVoiceReactNative (PreflightTest)
     }
     
     NSString *jsonSample = [self preflightStatsSampleToJsonString:self.preflightTest.latestSample];
+    if (jsonSample == nil) {
```
kpchoy: **[critical] "Unreachable today" contradicts the vendored nullable headers**

The claim that the native layer reports an all-zero sample "not as nil, so this path is unreachable today" is contradicted 100 lines below in this same file ("The sample is nil until the preflight test generates its first one", line 174; "The report is nil until one is available", line 213) and by the pinned TwilioVoice 6.13.6 headers:

```objc
// TVOPreflight.h
@property (nonatomic, strong, readonly, nullable) TVOPreflightReport *preflightReport;    // line 55
@property (nonatomic, strong, readonly, nullable) TVOPreflightStatsSample *latestSample;  // line 60
```

So on iOS the routine pre-first-sample state reaches this guard and rejects with `InvalidStateError` carrying a "could not serialize" message — a normal state misreported as a serialization defect.

The claim has been propagated into five places: `src/PreflightTest.ts` public TSDoc, `src/type/NativeModule.ts:140` ("this always resolves with a JSON string"), `PreflightTestModuleProxy.java:80,98`, `CHANGELOG.md:38` ("There is no separate absent state to detect"), and `src/__mock-data__/PreflightTest.ts:270` ("Both platforms report **this exact** all-zero-valued sample/report"), where it is hardened into a fixture. The new tests mock that payload, so they confirm the assumption rather than test it — and consumers are told explicitly not to handle a rejection iOS will produce.

Found by 4 agents (comment-analyzer, silent-failure-hunter, docs-validator, language-pattern-reviewer).

**Suggestion:** Verify on-device what `latestSample`/`preflightReport` return before the first sample. Then either have iOS synthesize the all-zero payload so the documented cross-platform contract holds, or keep the reject and correct the TSDoc, `NativeModule.ts`, and CHANGELOG — with a message that names the state ("has not generated a sample yet") rather than blaming serialization. Either way drop the "unreachable today" comments and the "this exact" fixture claim.

kpchoy (commented):
n/a

src/AudioDevice.tsx
```diff
@@ -90,6 +90,16 @@ export class AudioDevice {
...
+    // failure is undetectable. The native layer reports a failure by resolving
+    // with a rejection envelope rather than by rejecting, see
+    // `TwilioVoiceReactNative+PromiseAdapter.m` and
+    // `ReactNativeArgumentsSerializer.serializePromiseErrorWithName`, so
+    // awaiting the native promise directly discards it. Both platforms have a
+    // real failure path: `VoiceModuleProxy.selectAudioDevice` rejects on an
+    // unknown uuid, and `TwilioVoiceReactNative.m` rejects when the port lookup
+    // fails. Wrapping this in `settleNativePromise` fixes it, which matches the
+    // behavior this method is already documented to have.
     await NativeModule.voice_selectAudioDevice(this.uuid);
```
kpchoy: **[high] select() still swallows every native failure**

This awaits the raw `NativePromise` envelope instead of unwrapping it with `settleNativePromise`, so every failure resolves successfully — while the TSDoc three lines above still promises "Rejects if the audio device cannot be selected."

Both platforms have live failure paths: `VoiceModuleProxy.selectAudioDevice:119-124` rejects on an unknown uuid, and `ios/TwilioVoiceReactNative.m:713-723` rejects when the port lookup fails. The Android path is routinely reachable — `AudioSwitchManager.start()` mints a fresh `UUID.randomUUID()` per device on every route change, so any handle held across a headset plug/unplug has a dead uuid.

This PR applies the exact one-line fix to `CallInvite.reject` in the same diff. Leaving it deferred also makes the new `AudioDevice.Type.WiredHeadset` undetectably broken on iOS: the new `selectAudioDevice:` wired-headset branch is documented by its own comment (VBLOCKS-7140) as unable to succeed, and `select()` reports that permanent failure as success.

Found by 5 agents (bug-hunter, code-reviewer, silent-failure-hunter, type-design-reviewer, comment-analyzer).

**Suggestion:** `await settleNativePromise(NativeModule.voice_selectAudioDevice(this.uuid));`, add the rejection test (the new `CallInvite.reject` test is the template), and drop the comment — the existing TSDoc then becomes true. If the deferral is deliberate, correct the TSDoc in this PR so the published contract is not false for the life of VBLOCKS-7139.

ios/TwilioVoiceReactNative+PreflightTest.m
```diff
@@ -302,10 +347,18 @@ - (void)sendPreflightEvent:(NSDictionary *)eventPayload {
 }
 
 - (void)preflight:(nonnull TVOPreflightTest *)preflightTest didCompleteWithReport:(nonnull TVOPreflightReport *)report {
+    NSString *jsonReport = [self preflightReportToJsonString:report];
+    // A nil value in a dictionary literal raises an NSInvalidArgumentException.
+    // The report is nonnull here, so only a serialization failure reaches this.
+    if (jsonReport == nil) {
+        NSLog(@"Dropping PreflightTest completed event, could not serialize the report.");
```
kpchoy: **[high] Dropped terminal Completed event leaves consumers hung**

Returning early here means the JS `PreflightTest` emits neither `Completed` nor `Failed`. `didCompleteWithReport:` is terminal — the SDK will not call `didFailWithError:` afterwards — so anything awaiting a terminal event waits forever with nothing on the JS side. The previous behavior (`NSInvalidArgumentException`) was also wrong but loud; this converts a crash into a permanent silent hang.

The trigger is reachable: `dictionaryReport` carries `double` members (`mos`, `packetsLostFraction`, and the `TVOPreflightAggregateStats` average/min/max triples), and NaN or infinity — what a degenerate run with zero samples produces — makes serialization fail.

The same early return for `didReceiveStatsSample:` is fine; samples are periodic and non-terminal.

Found by 3 agents (silent-failure-hunter, bug-hunter, code-reviewer).

**Suggestion:** Emit `kTwilioVoiceReactNativePreflightTestEventTypeValueFailed` with the error payload that event already carries, so the JS layer reaches a terminal state, and propagate the underlying `NSError` into the message.

ios/TwilioVoiceReactNative+PreflightTest.m
```diff
@@ -173,18 +201,35 @@ - (NSString *)preflightStatsSampleToJsonString:(TVOPreflightStatsSample *)statsS
...
+
+    NSDictionary *reportDict = [report dictionaryReport];
+    if (reportDict == nil) {
+        return nil;
+    }
+
     NSError *jsonParseError;
-    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:[report dictionaryReport] options:NSJSONWritingFragmentsAllowed error:&jsonParseError];
+    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:reportDict options:NSJSONWritingFragmentsAllowed error:&jsonParseError];
     if (jsonParseError != nil) {
```
kpchoy: **[high] NSJSONSerialization raises; it does not return an NSError here**

Both hardened serializers treat `+dataWithJSONObject:options:error:` as error-returning. For an object that is not a valid JSON object it **raises `NSInvalidArgumentException`** instead — the `NSError **` out-parameter covers only a narrow set of write failures.

So the new `if (jsonData == nil)` guards cannot be reached for the failure mode that matters: NaN or infinity in the SDK-controlled doubles (`mos`, `packetsLostFraction`, the aggregate-stats triples). The process crashes before either check runs, which means the new nil-guards give false confidence — the comments say a serialization failure now surfaces as a rejected promise or a dropped event, but a real one still crashes the app.

**Suggestion:** Guard with `+[NSJSONSerialization isValidJSONObject:]` (and/or `@try`) before serializing, in both `preflightStatsSampleToJsonString` and `preflightReportToJsonString`, so the real failure routes into the `nil` return the callers were just taught to handle.

CHANGELOG.md
```diff
@@ -23,6 +101,8 @@
 
 ## Fixes
 
+- `CallInvite.reject` now rejects when the native layer reports an error. Previously the returned `Promise` resolved regardless of the outcome.
```
kpchoy: **[high] CallInvite.reject fix claim is untrue on iOS**

The JS wrap in `settleNativePromise` is correct, but it only has effect where native produces a rejection envelope — and this PR's own diff adds the acknowledgement that iOS does not:

```objc
// VBLOCKS-7159
// This method resolves the JS promise unconditionally.
```

There is no lookup and no failure path there, so rejecting a stale, unknown, or already-ended invite still silently "succeeds" on iOS. On Android the only rejecting path is `getCallRecord`'s pre-call guard (`InvalidArgumentError` for a missing/expired record), not "the native layer reports an error" during rejection. Every other platform-conditional behavior in this CHANGELOG is filed under "Platform Specific Fixes".

Found by 3 agents (docs-validator, silent-failure-hunter, comment-analyzer).

**Suggestion:** Either fix iOS `callInvite_reject` to look up the invite and reject (the `getCallInviteFromMap:` guard directly below is the shape), or move this entry under Platform Specific Fixes → Android and note that iOS still resolves unconditionally pending VBLOCKS-7159. Also add the newly possible rejection to `CallInvite.reject()`'s TSDoc, which still documents only the resolve.

test/appium-harness/src/test-suites/rtc-stats.ts
```diff
@@ -0,0 +1,425 @@
...
+
+      log.info(JSON.stringify({ localAudioTrackCount: trackCount }));
+
+      expect(
+        trackCount > 0,
+        'the call reported at least one local audio track',
+      ).toBe(true);
+    },
+  },
+  // KNOWN FAILING on both platforms: RTCStats.IceCandidatePairState holds
```
kpchoy: **[high] New e2e suites can never report success**

Three new suites embed steps documented as deterministically failing against defects this PR does not fix, yet each computes `setTestStatus(failed === 0 ? 'success' : 'failure')`:

- this step, `ice-candidate-pair-state-is-a-known-value` (VBLOCKS-7160 — `IceCandidatePairState` values never match native)
- `audio-device.ts:226` `select-wired-headset` (iOS, VBLOCKS-7140) and `:284` `select-a-removed-device-rejects` (both platforms, VBLOCKS-7139)
- `call-invite-reject.ts:254` `second-reject-rejects` (iOS, VBLOCKS-7159)

Because `summarizeSuiteResults` treats any non-`success` outcome as failed, both new aggregate runners — `unattended-all` and `attended-all` — are permanently red. With no green baseline, a genuine regression is indistinguishable from the known failures.

**Suggestion:** Add an explicit expected-failure marker (`knownFailing: { ticket, platforms }`) that `runSteps` excludes from the `failed` count driving `setTestStatus`. That also makes a suite go red when a known-failing step starts *passing*, which is the signal the ticket is fixed.

src/type/RTCStats.ts
```diff
@@ -31,9 +31,9 @@ export namespace RTCStats {
     relayProtocol: string;
     remoteCandidateId: string;
     remoteCandidateIp: string;
-    requestsReceieved: number;
+    requestsReceived: number;
```
kpchoy: **[high] IceCandidatePairState values never match native output**

This PR fixes three `RTCStats` keys that never matched native. The identical gap sits one type away in this same file and is left in place: `IceCandidatePairState` declares `STATE_FAILED = 'STATE_FAILED'`, `STATE_SUCCEEDED = 'STATE_SUCCEEDED'`, etc., while both platforms emit the camelCase constants — `ios/Utilities/TwilioVoiceStatsReport.h:156` and `android/.../StatsListenerProxy.java:164` both return `constants.src`'s `stateFailed` / `stateSucceeded` / …

So `stats.state === RTCStats.IceCandidatePairState.STATE_SUCCEEDED` is always false, on every platform. iOS additionally emits `stateUnknown`, which has no enum member at all.

These literals are published in `api/voice-react-native-sdk.api.md`, so fixing it later is another breaking change. This release already takes the breaking hit on `RTCStats` field names and `Call.QualityWarning` values in the same CHANGELOG section — one more bullet now versus another major bump later.

Found by 2 agents (type-design-reviewer, test-analyzer).

**Suggestion:** Redefine the members from `Constants.State*` the way `Call.QualityWarning` now is, add `STATE_UNKNOWN`, mirror the `Call.test.ts` constant-agreement test pair, and add a CHANGELOG entry. Decide whether Android should also emit `StateUnknown` (its switch has no such arm).

src/PreflightTest.ts
```diff
@@ -482,10 +482,26 @@ export class PreflightTest extends EventEmitter {
...
       common.NativeModule.preflightTest_getEndTime(this._uuid)
-    ).then(Number);
+    );
+
+    if (nativeEndTime === null || typeof nativeEndTime === 'undefined') {
+      return undefined;
+    }
+
+    const endTime = Number(nativeEndTime);
+
```
kpchoy: **[high] getEndTime reports malformed native values as "has not ended"**

`!Number.isFinite(endTime) || endTime === 0` folds two unrelated conditions together. `endTime === 0` is the documented "has not ended" signal and is correct. `!Number.isFinite(endTime)` means native sent something that is not a number — a native defect — now reported as the normal "still running" state, so a consumer polling for completion polls forever with nothing raised or logged. The TSDoc still says "Rejects if the native layer encountered an error."

This is the same conflation the PR removes elsewhere (iOS `getDeviceToken` no longer returns `""`, Android `showAvRoutePickerView` no longer no-ops), and the PR's own VBLOCKS-7141 comments argue malformed native payloads should surface via `constructInvalidValueError` — which this file uses nine times.

Separately, the `nativeEndTime === null || typeof nativeEndTime === 'undefined'` guard is unreachable against `preflightTest_getEndTime(): NativePromise<string>` (`src/type/NativeModule.ts:138`). Either the branch is dead or the native type should be `NativePromise<string | null>`; the new test passes `null` through an `any`-typed mock, so it pins neither.

Found by 2 agents (silent-failure-hunter, type-design-reviewer).

**Suggestion:** Split the branches — return `undefined` for `0`, throw `constructInvalidValueError` for a non-finite value — and reconcile the null guard with the declared native type.

src/Voice.tsx
```diff
@@ -752,24 +759,19 @@ export class Voice extends EventEmitter {
...
    * @returns
    * A `Promise` that
    *  - Resolves when the AV Route Picker View is shown.
+   *  - Rejects with an {@link TwilioErrors.UnsupportedPlatformError} when
+   *    invoked on any platform other than iOS.
    */
   async showAvRoutePickerView(): Promise<void> {
     switch (Platform.OS) {
       case 'ios':
         await settleNativePromise(NativeModule.voice_showNativeAvRoutePicker());
```
kpchoy: **[high] showAvRoutePickerView resolves on iOS even when the picker is never shown**

Removing the Android no-op is right, and the contract is now "Resolves when the AV Route Picker View is shown." iOS does not honor it (`ios/TwilioVoiceReactNative.m:725-745`): when `rootViewController` is nil the picker is never presented and the promise still resolves — the same "failure indistinguishable from success" defect the CHANGELOG says this change fixes, still present on the only platform where the method is supported.

Presentation is also `dispatch_async`'d, so resolution does not mean "shown" even on the success path. `windows[0]` additionally raises `NSRangeException` on an empty windows array, reachable in scene-based apps during backgrounding (pre-existing).

**Suggestion:** Reject with `InvalidStateError` when no window or root view controller is available instead of falling through to `resolvePromise`, and guard `windows[0]` with a count check. If resolving before presentation is intended, weaken the TSDoc to "has been requested".

src/PreflightTest.ts
```diff
@@ -515,14 +532,18 @@ export class PreflightTest extends EventEmitter {
    * @returns
    * A Promise that
    * - Resolves with the final {@link (PreflightTest:namespace).Report}.
-   * - Resolves with `undefined` if the report is unavailable.
+   * - Resolves with an all-zero-valued `Report` before the PreflightTest is
+   *   ready, since that is what the native layer itself reports for "not
+   *   ready" on both platforms.
    * - Rejects if the native layer encountered an error.
    */
```
kpchoy: **[high] All-zero sentinel is typed as a valid Report/RTCSample**

`getEndTime` was widened to `Promise<number | undefined>` in this PR precisely because "no real end time" needed to be representable. Its two siblings keep `Promise<PreflightTest.Report>` / `Promise<RTCSample>` while the new fixtures spell out what they resolve with before the test is ready: `callSid: ''`, `edge: ''`, empty arrays, every `TimeMeasurement` zeroed. That value type-checks as a `Report` while satisfying none of a `Report`'s invariants, and TypeScript actively tells the consumer a check is unnecessary.

The CHANGELOG's answer — listen for `Event.Sample`/`Event.Completed` instead of inspecting the resolved value — puts a real invariant in prose on a GA API. It is also internally inconsistent: `getEndTime` translates the `"0"` sentinel because no real end time can be `0`, while `callSid: ''` and an all-zero `testTiming` are exactly as unambiguous and pass straight through.

**Suggestion:** Widen both to `| undefined` and detect the sentinel at the boundary (`if (!reportObj.callSid) return undefined;`, `timestamp === '0'` for the sample), or use a discriminated `{ status: 'ready' … } | { status: 'not-ready' }` that generalizes to `getEndTime`. Note this interacts with the nullable-header finding — settle the native behavior first.

android/src/main/java/com/twiliovoicereactnative/PreflightTestModuleProxy.java
```diff
@@ -40,8 +42,15 @@ private void getPreflightTest(
...
     mainHandler.post(() -> {
       logger.debug(String.format(".getPreflightRecord(%s) > runnable", uuidStr));
-      final Object result = onSuccess.apply(record.getPreflightTest());
-      promise.resolve(result);
+      // This runs on the main handler, so an uncaught throw from the native
+      // layer would crash the application instead of settling the promise.
+      try {
+        final Object result = onSuccess.apply(record.getPreflightTest());
+        promise.resolve(result);
+      } catch (RuntimeException error) {
```
kpchoy: **[high] Catch logs nothing at all when getMessage() is null**

`logger.warning(error.getMessage())` loses the exception type and stack trace, and for the most likely `RuntimeException` here — an NPE from the native layer — `getMessage()` is null. `SDKLog.warning(String)` then NPEs on `message.getBytes()` inside its own `catch (Exception ignore) {}`, so nothing is logged at all and JS receives an `InvalidStateError` with a null message. The catch added to prevent a main-thread crash turns a diagnosable crash into an invisible failure.

(Two other agents confirmed this does not itself crash — `constructJSMap` filters null values — so this is a diagnosability regression, not a stability one.)

Catching all of `RuntimeException` also relabels genuine bugs in this module's own lambdas, including the two `IllegalStateException`s this PR adds, as `InvalidStateError`.

**Suggestion:** Use the stack-trace overload `SDKLog` already provides and a non-null message:

```java
} catch (RuntimeException error) {
  logger.warning(error, String.format("getPreflightTest(%s) failed", uuidStr));
  final String message = (error.getMessage() != null) ? error.getMessage() : error.toString();
  promise.rejectWithName(CommonConstants.ErrorCodeInvalidStateError, message);
}
```

CHANGELOG.md
```diff
@@ -7,6 +7,84 @@
...
+
+### Call.QualityWarning
+
+- Added `Call.QualityWarning.ConstantAudioOutputLevel`. Both native SDKs report this warning, but this SDK did not describe it.
+
+- Renamed `Call.QualityWarning.HighPacketLoss` to `Call.QualityWarning.HighPacketsLostFraction`, and its value is now `"high-packets-lost-fraction"` instead of `"high-packet-loss"`. Both platforms now report that value.
+
+  - Applications that reference `Call.QualityWarning.HighPacketLoss`, or that compare against a hard-coded `"high-packet-loss"`, need to use `Call.QualityWarning.HighPacketsLostFraction` instead.
+
+- On iOS, a constant audio output level warning, and any warning that this SDK does not recognize, previously reported the value `"undefined"`. Constant audio output level warnings now report `Call.QualityWarning.ConstantAudioOutputLevel`, and unrecognized warnings are omitted rather than reported as a value that is not a member of `Call.QualityWarning`.
```
kpchoy: **[high] Android quality-warning omission documented as iOS-only**

Android changed the same way in this PR: `serializeCallQualityWarnings` no longer emits `warning.toString()` for every warning but maps through `callQualityWarningToString`, which returns null outside the six mapped cases, and those are dropped with a log. An Android consumer reading this bullet concludes only the value rename affects them.

The explicit mapping also means Android-reported values are now the `constants.src` strings rather than whatever `toString()` produced, and the CHANGELOG never says which Android strings changed beyond the packet-loss rename.

For the native-SDK-upgrade checklist: both `default:` branches are unreachable at the pinned versions (all six enum members are mapped — verified against `voice-android:6.10.3` and TwilioVoice 6.13.6), so nothing is dropped today. A seventh warning in a future bump would be silently omitted, and an all-unrecognized set would arrive as an empty array indistinguishable from "warnings cleared".

**Suggestion:** Move the omission sentence out of the iOS-only bullet ("on both platforms… previously `"undefined"` on iOS and the raw native string on Android") and list any other Android-reported value that changed.

android/src/main/java/com/twiliovoicereactnative/PreflightTestModuleProxy.java
```diff
@@ -40,8 +42,15 @@ private void getPreflightTest(
 
     mainHandler.post(() -> {
       logger.debug(String.format(".getPreflightRecord(%s) > runnable", uuidStr));
```
kpchoy: **[high] Native preflight crash fixes have no CHANGELOG entry**

Two crash fixes in this PR are undocumented, in a release whose Fixes section already itemizes an Android NPE crash fix:

- Android: this try/catch means a throw from `getLatestSample()`/`getReport()` no longer crashes the app; the `Promise` rejects instead.
- iOS: `preflight:didReceiveStatsSample:` / `didCompleteWithReport:` previously raised `NSInvalidArgumentException`; now the event is dropped. That changes what an application can rely on receiving — which matters because the `getLatestSample`/`getReport` CHANGELOG entry tells applications to rely on those events instead of the resolved value.

**Suggestion:** Add an Android entry ("`PreflightTest` methods no longer crash the application when the native layer throws; the `Promise` rejects with an `InvalidStateError` instead") and an iOS one describing the dropped-event behavior — ideally after resolving the terminal-event finding.

src/__tests__/Call.test.ts
```diff
@@ -762,6 +766,45 @@ describe('Call namespace', () => {
       expect(typeof Call.QualityWarning).toBe('object');
     });
 
+    /**
+     * The native layer serializes call quality warnings using these values, so
+     * a mismatch means a warning never matches this enumeration. See
+     * `warningNameWithNumber` in `TwilioVoiceReactNative.m` on iOS, and
+     * `Call.CallQualityWarning.toString()` on Android.
```
kpchoy: **[high] Comment cites the Android toString() path this PR deleted**

This docblock says to see "`Call.CallQualityWarning.toString()` on Android". This same PR replaced that with the explicit `callQualityWarningToString` switch, precisely because — as that function's own comment says — the platforms name the packet-loss warning differently below this layer.

A maintainer debugging a warning mismatch would look for an enum-name-to-string relationship, find that `WARN_HIGH_PACKET_LOSS` does not stringify to `high-packets-lost-fraction`, and could "fix" the constant to match `toString()` — reintroducing the divergence this PR removed.

**Suggestion:** Point at `ReactNativeArgumentsSerializer.callQualityWarningToString`, and note the Android mapping is explicit *because* the native enum's `toString()` does not match these values.

src/PreflightTest.ts
```diff
@@ -637,6 +658,16 @@ function parseState(nativeState: string): PreflightTest.State {
 
 /**
  * Parse a sample object and transform the keys to match the expected output.
+ *
+ * VBLOCKS-7141: this reads every member without validating it, so a partially
```
kpchoy: **[high] VBLOCKS-7141 comment states a false contrast**

The comment asserts "Unlike `parseReport`, which throws on a partial payload, this surfaces nothing at all." `parseReport` reads `callSid`, `edge`, `iceCandidates`, `networkStats`, `selectedEdge` and `selectedIceCandidatePair` with bare property access and no validation — a payload missing any of them yields `undefined` members and throws nothing, exactly like `parseSample`. It throws only for `networkTiming` and `statsSamples`, which are dereferenced a level deeper. The two differ in which subset blows up, not in kind.

Both VBLOCKS-7141 comments then prescribe a `constructInvalidValueError` fix in enough detail to implement, in a file that already imports and uses that helper nine times, and the fix is not applied.

**Suggestion:** Correct or drop the false contrast. Then either apply the prescribed validation, or reduce both comments to a bare observation and let the ticket hold the fix plan — the plan is what rots, the observation is not.

test/appium-harness/src/test-suites/unattended-all.ts
```diff
@@ -0,0 +1,202 @@
+import * as React from 'react';
```
kpchoy: **[high] Native-only changes have no automated coverage**

The repo has no native unit tests (no `*Test.java`, no XCTest target), and `test/appium-orchestrator/test-suites/index.mjs:22` hard-codes `await outgoingCallTest(testElements)` as the only entry point — so of the twelve harness suites, exactly one runs in CI. This PR adds `unattended-all`, the natural fix, but does not wire it in.

Changed native paths with no test that can fail anywhere in the pipeline: Android `callQualityWarningToString` (only the constant-audio warnings are provoked by the e2e TwiML), iOS `warningNameWithNumber`'s nil branch, iOS `audioPortTypeMapping` for `AVAudioSessionPortHeadphones`, the Android `AudioSwitchManager` WiredHeadset/Earpiece split, and the iOS `getDeviceToken` rejection (the `voice-api` step accepts either outcome, so it cannot distinguish the old contract from the new).

**Suggestion:** Wire `unattended-all` into the orchestrator (parameterize the suite id, defaulting to `unattended-all` in CI) — this depends on the known-failing-step fix, or CI is permanently red. The Android warning mapping is pure and platform-independent, so an `android/src/test` JVM test is cheap insurance for what is now the only thing between a native enum and the public `Call.QualityWarning` values.

ios/TwilioVoiceReactNative+PreflightTest.m
```diff
@@ -81,6 +91,16 @@ @implementation TwilioVoiceReactNative (PreflightTest)
     }
     
     NSString *jsonReport = [self preflightReportToJsonString:self.preflightTest.preflightReport];
+    if (jsonReport == nil) {
+        // The native layer reports "not ready" as an all-zero report, not as
```
kpchoy, line 95: **"Unreachable today" contradicts the vendored nullable headers**

The claim that the native layer reports an all-zero sample "not as nil, so this path is unreachable today" is contradicted 100 lines below in this same file ("The sample is nil until the preflight test generates its first one", line 174; "The report is nil until one is available", line 213) and by the pinned TwilioVoice 6.13.6 headers:
```
 // TVOPreflight.h
 @property (nonatomic, strong, readonly, nullable) TVOPreflightReport *preflightReport;    // line 55
 @property (nonatomic, strong, readonly, nullable) TVOPreflightStatsSample *latestSample;  // line 60
```
So on iOS the routine pre-first-sample state reaches this guard and rejects with InvalidStateError carrying a "could not serialize" message, a normal state misreported as a serialization defect.

CHANGELOG.md
kpchoy, line 104: I think this is Android-only in effect. iOS `callInvite_reject` is unchanged, and this
PR adds the comment saying so ("This method resolves the JS promise unconditionally",
VBLOCKS-7159).

On Android the only rejecting path I can find is the `getCallRecord` guard for a
missing/expired record, which makes this "rejecting an invite whose record is already
gone now rejects" rather than "rejects when the native layer reports an error." Is
there another settle site I'm missing?

Could we scope the entry and move it under Platform Specific Fixes → Android? Also,
`CallInvite.reject()`'s TSDoc still documents only the resolve.

android/src/main/java/com/twiliovoicereactnative/PreflightTestModuleProxy.java
```diff
@@ -40,8 +42,15 @@ private void getPreflightTest(
...
-      final Object result = onSuccess.apply(record.getPreflightTest());
-      promise.resolve(result);
+      // This runs on the main handler, so an uncaught throw from the native
+      // layer would crash the application instead of settling the promise.
+      try {
```
kpchoy, line 47: Worth a CHANGELOG entry? Previously a throw from `getLatestSample()`/`getReport()`
crashed the app on the main handler, and now the `Promise` rejects instead. The Fixes
section already itemizes the CallInvite NPE crash fix, so it'd be consistent to list
this one.

Same on the iOS side: `didReceiveStatsSample:`/`didCompleteWithReport:` used to raise
`NSInvalidArgumentException` and now drop the event instead. That changes what an app
can rely on receiving, which matters because the `getLatestSample`/`getReport` entry
points people at those events.

kpchoy (changes requested):
Overall very thorough PR, left some comments

charliesantos:
I'm ok to get this merged in after a Kevin's +1

android/src/main/java/com/twiliovoicereactnative/PreflightTestModuleProxy.java
mhuynh5757: Neither path is reachable, so there's no behavior an app can observe changing.

On Android, `PreflightTest.getLatestSample()`/`getReport()` catch `JSONException` and return `new JSONObject()` (`PreflightTest.java:185-193`, `:214-222`), so neither returns null, and the `ThreadChecker` throw is gated on `com.twilio.voice.BuildConfig.DEBUG`, baked to false in the published `com.twilio:voice-android` AAR.

On iOS the delegate arguments are always constructed non-nil (`PreflightObserverAdapter.h:37`, `:90-91`), and the only two doubles in the sample dictionary are guarded in the core (`packets_lost_fraction` divide-by-zero guard at `call_impl.cpp:1271-1273`, `mos` range-clamped at `stats_utils.cpp:48-49`), so `NSJSONSerialization` can't fail either.

These guards keep the declared types honest if the native contract changes; they don't fix a crash anyone could hit at this point in time.

ios/TwilioVoiceReactNative+PreflightTest.m
mhuynh5757: I'll discuss this one with you over slack 👍

kpchoy (approved):

PR #721, opened 2026-09-14
feat: add Expo support via a config plugin and restore bare React Native (1.8.0)

charliesantos:
Supersedes the `2.0.0-preview.x` line. That approach delivered Expo support by
converting the Android binding to the Expo Modules API, which required disabling
React Native autolinking and so dropped bare React Native support and forced a
major version. This ships the same capability additively as 1.8.0.

## What changed

- **Expo**: new config plugin (`app.plugin.js` -> `plugin/withTwilioVoice.js`).
  Expo apps add `"@twilio/voice-react-native-sdk"` to `plugins` and run
  `expo prebuild`. Development build required; Expo Go is not supported.
- **Bare React Native**: removed `react-native.config.js`, which had set
  `android: null` and disabled autolinking. Bare apps upgrade with no code change.
- Removed the Expo Modules binding (`ExpoModule`, `ExpoPackage`, the two lifecycle
  listeners, `expo-module.config.json`). The Android binding stays a standard RN module.
- `iceServers` and `iceTransportPolicy` on `Voice.connect` and `CallInvite.accept`.
- `AudioDevice.nativeType`.
- Native SDK bumps: Android `6.7.1` -> `6.10.3`, iOS `6.13.3` -> `6.13.6`.
- Native method errors are now consistently `TwilioErrors` subclasses. `message`
  and `code` are unchanged; apps reading the bridge's undocumented `error.userInfo`
  from methods like `Call.mute` should read `error.code` / `error.message`.

## Testing

`test/MASTER_TEST_PLAN.md` is added as the canonical procedure, with
`test/MASTER_TEST_CASES.md` and `test/API_TEST_CASES.md` as the case catalogs, plus
runner scripts for Android emulators, the iOS simulator, and iOS hardware.

Run results and open findings are in MASTER_TEST_PLAN sections 9 and 10. Summary:

- Expo and bare runtimes verified equivalent on both platforms.
- Expo config plugin builds verified on Expo SDK 52, 54, 55, 56, 57.
- Unit suites: 20 suites, 598 tests, 100% coverage.
- Incoming calls are Blocked on the current signing setup, not on missing coverage.
- Four e2e failures remain open (sections 9.5a-9.5c). A regression audit against
  tag `1.7.0` (section 9.7) confirms none were introduced by this branch.

Full changelog in `CHANGELOG.md`.


ios/TwilioVoiceReactNative+CallKit.m
```diff
@@ -205,6 +205,21 @@ - (void)makeCallWithAccessToken:(NSString *)accessToken
...
+            // `Voice.connect()` stayed pending for the lifetime of the app and
+            // the caller saw a hang rather than an error. Cleared afterwards so
+            // a later call cannot settle against a stale block.
+            if (self.callPromiseResolver) {
+                [self rejectPromiseWithName:self.callPromiseResolver
+                                       name:kTwilioVoiceReactNativeErrorCodeInvalidStateError
+                                    message:[NSString stringWithFormat:
+                                             @"CallKit rejected the start-call transaction: %@",
+                                             [error localizedDescription]]];
+                self.callPromiseResolver = nil;
```
kpchoy: **Promise resolver has no settle-once owner**

`voice_connect_ios` assigns `self.callPromiseResolver` *after* `makeCallWithAccessToken:` has already started the async transaction (`ios/TwilioVoiceReactNative.m:688`), and no `methodQueue` is declared — so this block can run first, observe nil, and leave `Voice.connect()` pending forever, which is the defect it exists to fix. The success path at :268 also resolves without nilling, so a later failure can double-invoke a settled block. And `= nil` here is an unsynchronized write to a `nonatomic` property the main queue reads.

**Suggestion:** Assign the resolver before starting the transaction, and take-and-clear it on one queue, keyed by the call UUID. That also closes the nil-`call` arm at :262.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
+const {
+  withMainApplication,
+  withMainActivity,
+  withInfoPlist,
+  withEntitlementsPlist,
+  createRunOncePlugin,
+} = require('@expo/config-plugins');
```
kpchoy: **Undeclared @expo/config-plugins breaks pnpm and Yarn PnP**

`@expo/config-plugins` is in neither `dependencies`, `peerDependencies`, nor `devDependencies` — it resolves only because `expo` hoists it, which is why no in-repo path catches this. pnpm's strict linker and Yarn PnP fail with MODULE_NOT_FOUND loading `app.plugin.js`, aborting `expo prebuild`.

**Suggestion:** Use `require('expo/config-plugins')` — same module, resolved through the already-declared optional `expo` peer. Don't add `@expo/config-plugins` to `dependencies`; that risks a second copy mismatched against the app's `expo`.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
...
+    modes.add('voip');
+    cfg.modResults.UIBackgroundModes = Array.from(modes);
+    return cfg;
+  });
+}
+
+function withVoiceEntitlements(config, { apsEnvironment }) {
+  return withEntitlementsPlist(config, (cfg) => {
+    cfg.modResults['aps-environment'] =
+      cfg.modResults['aps-environment'] || apsEnvironment;
```
kpchoy: **aps-environment silently defaults to development**

`existing || apsEnvironment`, with `apsEnvironment` defaulted to `'development'` at :305 — the no-props shape every doc example uses. A release build then registers its VoIP token against the APNs sandbox: the build succeeds, `Voice.register()` resolves, and incoming calls never arrive.

**Suggestion:** Warn when defaulting so it reaches prebuild output, or honour `config.ios?.entitlements?.['aps-environment']` first.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
...
+
+function withVoiceMainActivity(config) {
+  return withMainActivity(config, (cfg) => {
+    if (cfg.modResults.language !== 'kt') {
+      throw new Error(
+        `${TAG}: MainActivity must be Kotlin (found ${cfg.modResults.language}).`
+      );
+    }
+    let src = cfg.modResults.contents;
+    if (src.includes('voiceActivityProxy')) {
```
kpchoy: **Idempotency guard no-ops on a partially wired file**

`src.includes('voiceActivityProxy')` can't distinguish fully wired from partially wired. Given a MainActivity with the proxy field and the `onCreate` hook but neither override, the mod returns the file unchanged — no overrides, no `Intent` import, no error — and it compiles. `docs/expo/app-config.md:130-146` tells developers to hand-wire all three members, so this is reachable. The `voiceApplicationProxy` guard at :79 has the same hole.

**Suggestion:** Check for each member the plugin owns, and throw naming any that are missing.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
...
+
+const TAG = 'twilio-voice-react-native';
+const pkg = require('../package.json');
+
+/* ------------------------------------------------------------------ utils */
+
+function addImports(src, imports) {
+  let out = src;
+  for (const imp of imports) {
+    if (out.includes(imp)) continue;
```
kpchoy: **Import dedupe is a substring test**

`out.includes(imp)` matches substrings, and `import android.content.Intent` is a prefix of `import android.content.IntentFilter`. With IntentFilter already imported, the `Intent` import is skipped while `onNewIntent(intent: Intent)` is still emitted: prebuild succeeds, then Gradle fails on an unresolved reference with nothing naming this plugin.

**Suggestion:** Compare trimmed whole lines rather than substrings.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
...
+
+/** Insert `body` as the first statement of `fun name(...) { ... }`. */
+function insertAtStartOfMethod(src, signatureRegex, body, anchorAfter) {
+  const m = src.match(signatureRegex);
+  if (!m) return null;
+  const braceIdx = src.indexOf('{', m.index + m[0].length - 1);
+  if (braceIdx === -1) return null;
+  let insertAt = braceIdx + 1;
+  if (anchorAfter) {
+    const rest = src.slice(braceIdx);
```
kpchoy: **onCreate anchor search escapes the method body**

`rest = src.slice(braceIdx)` spans the rest of the file, and an unmatched `anchorAfter` silently falls back to `braceIdx + 1`. With an expression-bodied `override fun onCreate(...) = setup(...)`, the init call is injected into `setup`, where the parameter is out of scope. With no match at all it lands before `super.onCreate`, so the proxy calls `requestPermissions()` on an activity whose super hasn't run.

**Suggestion:** Bound the search to the matched method's braces, and throw on a missing anchor or an expression body.

plugin/withTwilioVoice.js
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * Insert `body` just before the final closing brace of the file's class.
+ *
+ * Throws rather than returning null. Callers assign the result straight to
+ * `modResults.contents`, so a null return would blank the generated file and
+ * surface later as an unrelated Kotlin syntax error.
+ */
+function insertBeforeLastBrace(src, body) {
+  const i = src.lastIndexOf('}');
```
kpchoy: **insertBeforeLastBrace injects into a trailing declaration**

`src.lastIndexOf('}')` assumes the class brace is the file's last. With a trailing top-level `fun helper(): Int { return 1 }`, `onTerminate()` is injected inside `helper()` after its `return` — uncompilable Kotlin, silently.

**Suggestion:** Find the class body's matching close brace instead, and throw if it can't be identified.

android/src/main/java/com/twiliovoicereactnative/AudioSwitchManager.java
```diff
@@ -77,6 +77,20 @@ public static String getAudioDeviceNativeType(AudioDevice audioDevice) {
...
+   * Identity for an AudioDevice that survives AudioSwitch handing back a new
+   * instance for the same physical device.
+   *
+   * Uses the type string from {@link #getAudioDeviceNativeType} rather than the
+   * class name, for the same reason that method does: code shrinkers rename
+   * AudioSwitch's classes in a consuming app's release build. The name is
+   * included so two devices of the same type, such as two Bluetooth headsets,
+   * keep distinct handles.
+   */
+  private static String audioDeviceIdentity(AudioDevice audioDevice) {
```
kpchoy: **Device identity key collides**

`type|name` collides for two same-model Bluetooth headsets. The loop reads the persistent map but writes a local `seen`, replacing the persistent map only afterwards — so the first callback gives each a distinct UUID, and every later one maps both to the single stored UUID, collapsing the list to N-1. `select()` triggers a callback, so the list shrinks right after a selection, and the dropped UUID becomes a dead handle. Regression from the per-device random UUIDs.

**Suggestion:** Add an occurrence ordinal to the key, or mint a fresh UUID on collision and assert the count matches `devices.size()`.

src/Voice.tsx
```diff
@@ -324,7 +335,9 @@ export class Voice extends EventEmitter {
       this._handleNativeEvent
     );
 
-    NativeModule.voice_setExpoVersion(getExpoVersion());
+    this._expoVersionPromise = settleNativePromise(
```
kpchoy: **The new Expo-version gate misses both inbound-call paths**

Two paths bypass it. `CallInvite.accept` calls native with no await (`src/CallInvite.tsx:470`). And Android's inbound push runs no JS at all: on a cold start `VoiceFirebaseMessagingService.onMessageReceived` reaches `Voice.handleMessage` and produces insights before JS constructs `Voice`, so the property was never set — no await can cover that path.

**Suggestion:** Share the promise from `src/common.ts` so `CallInvite` can await it. For the push path, have the config plugin write the version into a Gradle `buildConfigField` and an Info.plist key, read in `VoiceApplicationProxy.onCreate`.

src/Voice.tsx
```diff
@@ -324,7 +335,9 @@ export class Voice extends EventEmitter {
...
+      NativeModule.voice_setExpoVersion(getExpoVersion())
```
kpchoy: **Synchronous throw escapes the catch**

The argument is evaluated before `.catch` is attached, so only async failures are swallowed — a synchronous throw propagates out of `new Voice()`. With the native module unresolved that's a `TypeError` from the constructor, contradicting the privateRemarks at :291-298 that this "must never prevent a call, registration, or preflight test".

**Suggestion:** Wrap the call: `(async () => settleNativePromise(NativeModule.voice_setExpoVersion(getExpoVersion())))().catch(() => undefined)`.

src/common.ts
```diff
@@ -5,22 +5,18 @@
  * See LICENSE in the project root for license information.
  */
 
-import { requireNativeModule } from 'expo-modules-core';
 import * as ReactNative from 'react-native';
 import type { TwilioVoiceReactNative as TwilioVoiceReactNativeType } from './type/NativeModule';
 import { getExpoVersion } from './utility/expoVersion';
 
-export const Platform = ReactNative.Platform;
+export const NativeModule = ReactNative.NativeModules
```
kpchoy: **Unresolved native module fails without a diagnostic on Android**

The unchecked cast leaves a missing module as `undefined` typed as present, and `new NativeEventEmitter(undefined)` is silent on Android — RN gates its invariant on iOS and skips the warning branch for `undefined`. The first symptom is a bare `TypeError` from the `Voice` constructor, where the removed `requireNativeModule` threw a named error. The new `common.test.ts` always stubs a present module, so this path is untested.

**Suggestion:** Throw a message naming the cause — unlinked module, run `pod install`, rebuild, Expo Go unsupported — before constructing the emitter.

CHANGELOG.md
```diff
@@ -1,61 +1,129 @@
...
+
+  - Twilio Voice iOS SDK upgraded from `6.13.3` to `6.13.6`.
+
+- Errors raised by native methods are now consistently surfaced as
+  `TwilioErrors` subclasses. Previously this applied to some methods but not
+  others: `Voice.connect` and `CallInvite.accept` already constructed a typed
+  error, while methods such as `Call.mute`, `Call.disconnect` and
+  `Voice.getVersion` let the underlying React Native bridge error propagate.
+  `message` and `code` are unchanged. Applications that read the bridge's
+  undocumented `error.userInfo` property from those methods should read
```
kpchoy: **Migration note is wrong about error.code**

The note says `code` is unchanged and tells apps to read `error.code` instead of `userInfo`. On the name-based path these methods take, `settleNativePromise` throws `InvalidStateError` / `InvalidArgumentError` / `UnexpectedNativeError`, all of which call `super(message)` with no code — so `error.code` is `undefined`, where 1.7.0 gave `'EUNSPECIFIED'`. The remedy the note offers doesn't exist.

**Suggestion:** Name the three thrown classes, state that `code` is `undefined` on them, and point consumers at `error.message` and `instanceof`.

kpchoy (changes requested):
## Review Summary

Two clusters:

- **iOS promise resolver** (`+CallKit.m:221`) — the new failure block can leave `connect()` hanging,
  settle a stale resolver, or over-release one. Critical; one fix closes all three.
- **Expo config plugin** — five defects, four of them silent: an undeclared dependency that breaks
  pnpm and Yarn PnP, `aps-environment` defaulting to `development`, and three source-patching
  helpers that mis-inject without erroring.

Also: the new Expo-version gate misses both inbound-call paths, an unlinked native module now fails
without a diagnostic on Android, the audio-device identity key collides, and the CHANGELOG points
consumers at `error.code`, which is `undefined` on the errors these methods now throw.

kpchoy (approved):
LGTM!

PR #723, opened 2026-09-15
[VBLOCKS-6611] Test Bare and Expo Appium Harness in CI

mhuynh5757:
## Submission Checklist

## Summary

Runs the Appium suites against real Sauce Labs devices in CI. Every push builds the Appium harness app and the bare example app for both platforms, uploads each build to Sauce Labs, and runs the orchestrated suites.

## What changed

- Twelve jobs in `ci.yml`: four builds, four Sauce Labs uploads, four e2e runs, covering the Expo harness and the bare example app on iOS and Android.
- Three composite actions the jobs share: `e2e-credentials` (OIDC exchange for Sauce credentials), `saucelabs-upload`, and `run-e2e-suites`.
- `.github/fixtures/` holds placeholder `google-services.json` files and a `secrets.json` template. The placeholders are why `registration-test` is not in the allowlist.
- Audio device suites: `select-audio-device` in `voice-api-test` now asserts only what a suite with no active call can assert. The readback assertion moved to a new `select-audio-device-while-connected` step in `call-controls-test`, which holds a call open so the iOS audio session is active.
- `show-av-route-picker-view` is Android only. The iOS picker is modal and cannot be dismissed, so XCUITest can no longer read the harness status.
- `prepare-release.yml` creates the version tag at prepare time, so release candidates get a tag without being published.
- `postcss` pinned to `8.5.28` via `resolutions`, covering a CVE pulled in by the Expo dev dependency.

## Test evidence

Full CI run showing green on Bare and Expo on iOS and Android.

## Known gaps

None change CI behavior today.

- Three `KNOWN FAILING on Android` comments in `voice-api.ts` and `preflight.ts`. Left in place until VBLOCKS-7133 and VBLOCKS-7134 are picked up. Note that they pass in Sauce Labs and GHA CI. Revisit or close as not-needed when acting on the tickets.
- Nothing asserts on Android that a selected audio device reads back as selected. `select-audio-device-while-connected` is gated to iOS for VBLOCKS-7133.
- The `ice-test` line in the `run-e2e-suites` summary says `Blocked`. The CI run reports `Pass`.
- `release.yml` still passes `--target` to `gh release create`, which GitHub ignores now that the tag already exists.


.github/actions/e2e-credentials/action.yml
```diff
@@ -0,0 +1,171 @@
...
+          echo "::error::Failed to obtain a GitHub OIDC JWT"
+          exit 1
+        fi
+        echo "::add-mask::$GITHUB_JWT"
+
+        # 2. Exchange the JWT for the SauceLabs service account.
+        #    Each value is masked the moment the value is parsed, before the
+        #    failure branch below can print anything.
+        SAUCE_RESP=$(curl -sS "$VENDOR_URL" \
+          -H "Authorization: Bearer ${GITHUB_JWT}" \
```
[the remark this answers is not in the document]
mhuynh5757: Not found to be the case in action. CI works.

.github/actions/e2e-credentials/action.yml
```diff
@@ -0,0 +1,171 @@
...
+        # 3. Mint the access token the harness dials with. The ttl is the
+        #    vendor's 3600 second ceiling rather than the vendor's 300 second
+        #    default, which would very plausibly expire during device
+        #    allocation, before the harness ever dials. A ttl above the
+        #    ceiling is capped rather than rejected.
+        TOKEN_REQ=$(jq -nc --arg identity "ci-${PLATFORM}-${GITHUB_SHA}-$(date +%s%3N)" \
+          '{action: "mint-voice-token", identity: $identity, ttl: 3600}')
+
+        TOKEN_RESP=$(curl -sS "$VENDOR_URL" \
+          -H "Authorization: Bearer ${GITHUB_JWT}" \
```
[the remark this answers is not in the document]
mhuynh5757: Doesn't seem to be the case in action, the CI works.

.github/actions/saucelabs-upload/action.yml
```diff
@@ -0,0 +1,176 @@
...
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}" \
+          | jq -r '.value' || true)
+        if [ -z "$GITHUB_JWT" ] || [ "$GITHUB_JWT" = 'null' ]; then
+          echo "::error::Failed to obtain a GitHub OIDC JWT"
+          exit 1
+        fi
+        echo "::add-mask::$GITHUB_JWT"
+
+        VENDOR_RESP=$(curl -sS "$VENDOR_URL" \
+          -H "Authorization: Bearer ${GITHUB_JWT}" \
```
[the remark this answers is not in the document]
mhuynh5757: Doesn't seem to be the case in action, CI works.

test/appium-harness/src/test-suites/voice-api.ts
```diff
@@ -142,17 +142,37 @@ const STEPS: Array<Step<Voice>> = [
...
+  // report that same device back - lives in `call-controls-test` instead. On
+  // iOS `selectAudioDevice` only sets a preference on the `AVAudioSession`,
+  // and `selectedDevice` is recomputed from the route that the session
+  // actually settled on. With no call in progress the session is never
+  // activated, so the route never moves and selecting the earpiece reports the
+  // speaker back.
+  //
+  // KNOWN FAILING on Android: selecting regenerates every device uuid, which
+  // the uuid-stability assertion below reports. Restores the original
+  // selection either way.
```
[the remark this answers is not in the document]
mhuynh5757: Does not fail in CI in practice. Will revisit with the ticket noted in the comment.

package.json
```diff
@@ -187,5 +187,8 @@
     "@twilio/voice-errors": "1.8.0",
     "eventemitter3": "^4.0.7"
   },
+  "resolutions": {
+    "postcss": "8.5.28"
+  },
```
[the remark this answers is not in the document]
mhuynh5757: Nested resolutions don't fail CI. While it'd be nice to be consistent, I think that less overrides the better. WDYT @charliesantos?

.github/workflows/ci.yml
```diff
@@ -1,8 +1,10 @@
 name: CI
+
 on:
```
kpchoy, line 3: **No concurrency group for the new paid device jobs**

This workflow triggers on every `pull_request` event and now adds four jobs that each hold a real Sauce Labs device for up to 30 minutes, with no `concurrency` block. Each push to an open PR starts a fresh set while the previous set runs to completion, so a PR accumulates N billed device sessions and superseded runs interleave their step summaries.

**Suggestion:** Add `concurrency: {group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}, cancel-in-progress: ${{ github.event_name == 'pull_request' }}}` beside the `on:` block.

.github/workflows/ci.yml
```diff
@@ -40,15 +44,36 @@ jobs:
...
 
-      - run: yarn install --immutable
+      # Yarn 4 turns on hardened mode for pull requests on a public repo, which
+      # re-resolves every descriptor against the registry instead of trusting
+      # the lockfile. Artifactory serves a broken packument for proxy-addr, so
+      # that resolution fails. Scoped to the lockfiles that carry proxy-addr.
+      # Remove once SSCHELP-2959 is fixed.
+      - name: Install dependencies
+        env:
+          YARN_ENABLE_HARDENED_MODE: "false"
```
kpchoy, line 54: **Yarn hardened mode disabled install-wide, not for one descriptor**

All six `YARN_ENABLE_HARDENED_MODE: "false"` sites (54, 75, 151, 256, 408, 534) are new in this PR, since the base branch has none. The comment says the real problem is one descriptor, `proxy-addr`, whose packument Artifactory serves broken, but the env var disables lockfile re-resolution and checksum auditing for the entire install, in jobs that hold `id-token: write` and then execute installed code. The fork guard is what keeps this below critical.

**Suggestion:** Pin `proxy-addr` in `resolutions`, as the orchestrator already does for `axios`, and drop the env var from all six steps.

.github/workflows/ci.yml
```diff
@@ -59,3 +84,877 @@ jobs:
       - name: Remove Artifactory credentials
         if: always()
         run: rm -f ~/.npmrc ~/.yarnrc.yml
+
+  ###
+  # Build Android Expo artifact
+  ###
+  build-android-expo:
```
kpchoy, line 91: **Build and e2e jobs do not depend on check**

The four build jobs run in parallel with `check` rather than after it, and the upload and e2e jobs chain only off their own build. A PR that fails type-check, lint, or unit tests still burns four builds, four Sauce Labs uploads, and four real-device sessions, exercising the credential vendor eight times before anyone looks at it.

**Suggestion:** Add `needs: check` to `build-android-expo`, `build-ios-expo`, `build-android-bare`, and `build-ios-bare`. The existing chains then gate uploads and e2e transitively.

.github/workflows/prepare-release.yml
```diff
@@ -137,18 +137,25 @@ jobs:
...
       #
       # Deliberately not a force-push. That would discard whatever is on the
       # existing branch. If this becomes annoying, check for the remote branch
       # before the build rather than relaxing the push.
-      - name: Push release branch
+      - name: Push release branch and tag
         env:
           RELEASE_VERSION: ${{ inputs.release_version }}
-        run: git push origin "release/${RELEASE_VERSION}"
+        run: |
```
kpchoy, line 158: **Branch and tag pushed non-atomically**

`git push origin "release/${VERSION}" "refs/tags/${VERSION}"` updates the two refs independently, so the tag can land while the branch push is rejected. That is exactly the re-dispatch scenario the comment above describes. Since `release.yml` now consumes this tag, the result is a tag pointing at an unpublished commit, and recovery is the manual delete-both the comment documents.

**Suggestion:** Add `--atomic` to the push so both refs land or neither does.

package.json
```diff
@@ -187,5 +187,8 @@
     "@twilio/voice-errors": "1.8.0",
     "eventemitter3": "^4.0.7"
   },
+  "resolutions": {
+    "postcss": "8.5.28"
```
kpchoy: **postcss exact-pinned, and diverges from sibling project pins**

`"postcss": "8.5.28"` is an exact override rather than a floor, so no future postcss security patch arrives via a lockfile refresh or `yarn up`. `test/expo` and `test/appium-harness` already pin `8.5.26`, so the repo now carries two different "fixed" versions of the same package with nothing tying them together. The `resolutions` mechanism itself is right, since postcss enters only transitively via `@expo/metro-config` and a devDependency bump would not touch it.

**Suggestion:** Use `^8.5.28` at the root and bump both sibling projects to match in this PR, each with its own `yarn install`.

.github/actions/saucelabs-upload/action.yml
```diff
@@ -0,0 +1,181 @@
...
+        echo "::notice::Uploading as ${FILENAME}"
+
+        # --write-out appends the status code on its own trailing line, so the
+        # failure branches below can name the status. The body is parsed for
+        # the file id but is never printed.
+        #
+        # `|| true` so a transport failure is reported as HTTP 000 rather than
+        # aborting on curl's exit code with no explanation.
+        UPLOAD_RESP=$(curl -sS \
+          -u "${SAUCE_USERNAME}:${SAUCE_ACCESS_KEY}" \
```
kpchoy, line 149: **Sauce Labs access key passed on the curl command line**

`-u "${SAUCE_USERNAME}:${SAUCE_ACCESS_KEY}"` puts a long-lived service-account credential in argv for the duration of the multipart upload, the slowest call in the action, where any co-resident process on the self-hosted runner can read it via `ps` or `/proc/<pid>/cmdline`. `::add-mask::` protects the log, not the process table. Every other credential in these two new actions is deliberately kept out of argv.

**Suggestion:** Feed it over stdin with `printf 'user = "%s:%s"\n' ... | curl -K -`, or write a `--netrc-file` under the existing `umask 077`.

.github/actions/run-e2e-suites/action.yml
```diff
@@ -0,0 +1,206 @@
...
+      env:
+        PLATFORM: ${{ inputs.platform }}
+        LABEL: ${{ inputs.label }}
+        STORAGE_FILENAME: ${{ inputs.storage-filename }}
+        RESULTS_JSON: ${{ github.workspace }}/temp/results.json
+      run: |
+        set -euo pipefail
+
+        {
+          echo "### SauceLabs e2e (${LABEL})"
```
kpchoy, line 155: **results.json handling can fail a passing run and drop its caveats**

This step sets `set -euo pipefail` and then reads the file with two bare `jq` calls, so malformed or truncated JSON fails the step, and the job, even when every suite passed. It also loses the "Not exercised by this run" block inside the same brace group. The file is not a trusted artifact: the orchestrator only `console.log`s a failed `writeFileSync`, and the `mkdir -p` at line 139 is the one new `run:` block without `set -euo pipefail`, so an unwritable path is silently ignored and then misreported as "the orchestrator did not reach its report".

**Suggestion:** Validate once with `jq -e . "$RESULTS_JSON"` and let each read fall back with `|| true`, add `set -euo pipefail` to the run step, and make the orchestrator write failure an `::error::`.

.github/actions/run-e2e-suites/action.yml
```diff
@@ -0,0 +1,206 @@
...
+            echo "| --- | --- | --- | --- |"
+            # `detail` is escaped because the orchestrator joins failing steps
+            # with a literal " | " and reports a rejection as its raw error
+            # message. Either one would otherwise split the row into extra
+            # columns or end the table early, on the failure path where the
+            # detail matters most.
+            jq -r '
+              def cell: (. // "") | gsub("\\|"; "\\|") | gsub("\r?\n"; " ");
+              .results[]
+              | "| \(.id | cell) | \(.status | cell) | \(.seconds // "") | \(.detail | cell) |"
```
kpchoy, line 176: **Unmasked orchestrator detail written to the job summary**

This step pipes each suite's `detail` field into `$GITHUB_STEP_SUMMARY`, where `::add-mask::` does not apply, so content appended there renders verbatim. Per the comment two lines above, `detail` carries raw rejection text from a driver built out of `secrets.json` and from suites dialing with the `token.json` access token, so a failure echoing request configuration could disclose a credential on exactly the path where the column matters most.

**Suggestion:** Keep only `id`/`status`/`seconds` in the summary and log `detail` where masking applies, or extend the `cell` filter with a JWT-shaped `gsub` redaction.

.github/actions/run-e2e-suites/action.yml
```diff
@@ -0,0 +1,206 @@
...
+          echo ""
+
+          if [ "$PLATFORM" = "android" ]; then
+            echo "- \`registration-test\`: the build receives a placeholder google-services.json, so it has no real FCM push credential."
+          else
+            echo "- \`registration-test\`: the build is unsigned and carries no aps-environment entitlement, so it can obtain no PushKit VoIP token."
+          fi
+
+          echo "- \`ice-test\` valid-* variants: no TURN credentials are bundled at build time, so those variants skip and the suite reports Blocked."
+        } >> "$GITHUB_STEP_SUMMARY"
```
kpchoy, line 194: **Two excluded suites are missing from the coverage caveats**

`SUITES` allowlists 8 of the 12 suites registered in `test/appium-harness/src/app/index.tsx`. Of the four excluded, `registration-test` is covered by this block and `incoming-call-test-manual` is explained in the `SUITES` comment above, but `quality-warnings-test` and `incoming-ice-test` appear in neither the results table nor the gap list, so a green run reads as covering them. Because the list is hand-maintained alongside the allowlist, it will drift again the next time `SUITES` changes. The `ice-test` Blocked-vs-Pass mismatch you already noted has the same root cause.

**Suggestion:** Emit every registered suite into `RESULTS_JSON` with a `not-run` status when it falls outside `SUITES` and render the gap list from that, so an omission cannot silently survive.

test/appium-harness/src/test-suites/voice-api.ts
```diff
@@ -163,17 +183,28 @@ const STEPS: Array<Step<Voice>> = [
...
           log.info(JSON.stringify({
             selected: audioDevice.name,
             reported: selectedDevice?.name,
           }));
 
           expect(
-            selectedDevice?.uuid,
-            `the selected device after selecting "${audioDevice.name}"`,
-          ).toBe(audioDevice.uuid);
+            audioDevices
```
kpchoy, line 195: **Reworked select-audio-device has three paths that pass without asserting**

Separate from the Android readback gap already noted in the description, the rewritten step can report `passed` while checking nothing. All assertions sit inside `for (const audioDevice of initial.audioDevices)` with no non-empty precondition, so an empty device list, a plausible bridge or native-record failure and one of the failures this step exists to catch, skips every assertion. Both remaining assertions also hold when `select()` is a complete no-op, because an unchanged uuid set is trivially stable and the already-selected device is always a member. The `if (typeof selectedDevice !== 'undefined')` guard then removes even the membership check in the case where native reports nothing selected at all.

**Suggestion:** Assert `initial.audioDevices.length > 0` before the loop, and replace the `undefined` guard with an assertion that `selectedDevice` is defined, since a device list with nothing selected is a defect rather than a skip condition.

kpchoy (changes requested):
## Review Summary

Two credential-handling issues in the new composite actions, plus some robustness and cost gaps in the CI wiring. The committed fixtures are clean (zeroed project numbers, all-A API key, `PLACEHOLDER` team ID) and no real credentials are present. Items already covered in your known-gaps section are left out below.

- **Credentials:** the Sauce Labs key is passed in `curl` argv, readable via `ps` on the self-hosted runner. Orchestrator `detail` text is written to the job summary, which `::add-mask::` does not cover.
- **Supply chain:** Yarn hardened mode is switched off for six whole installs to work around one broken descriptor, in jobs holding `id-token: write`.
- **Robustness:** `results.json` handling can redden a run whose suites all passed, and drops the coverage caveats with it. The release branch+tag push is non-atomic.
- **Coverage honesty:** the reworked `select-audio-device` has three paths that pass without asserting anything. `quality-warnings-test` and `incoming-ice-test` are excluded by `SUITES` but missing from both the results table and the gap list.
- **Cost:** no concurrency group on the new paid device jobs, and builds don't gate on `check`.
- **Dependencies:** `postcss` is exact-pinned at 8.5.28 while sibling projects pin 8.5.26. The pin's CVE fix is unverified, since no audit tool or network was available, so confirm with `yarn npm audit` or osv-scanner.

One question rather than a finding: `appium-xcuitest-driver` reads as `^12.11.1` to `^11.0.0` in the diff, but `main` already carries `^11.0.0` in `dependencies`, so this converges the branch with what ships. Confirm that's deliberate, since `^11.0.0` resolves to exactly 11.0.0 while `setup.mjs` targets iOS `platformVersion: '26'`.

Not filed: the `ios/TwilioVoiceReactNative.m` change is comment-only and VBLOCKS-7216 already tracks the underlying silent success. Roughly 60 lines of OIDC/Sauce credential exchange are duplicated between `e2e-credentials` and `saucelabs-upload`, plus four copies of the install block in `ci.yml`. The CHANGELOG checkbox is unchecked.

.github/workflows/ci.yml
mhuynh5757: Let me give that a shot just to be sure, but the error seems to reside in the Artifactory mirror in that `proxy-addr` `2.x` is not served to the CI jobs outright. It's not that the dependency is out-of-date, it's that the the up-to-date packages are not downloadable with hardened mode on.

mhuynh5757: Just gave it a shot @kpchoy, unfortunately did not work. 😞 

I believe at this time, `YARN_ENABLE_HARDENED_MODE: "false"` is the only unblocker.

.github/actions/saucelabs-upload/action.yml
mhuynh5757: I'll dig into this. Gut reaction though is that if we're worried about malicious co-resident process reading secrets, then isn't the machine already breached since there's a malicious process running on the machine? WDYT @kpchoy @charliesantos?

.github/actions/e2e-credentials/action.yml
```diff
@@ -0,0 +1,175 @@
...
+        #    failure branch below can report anything.
+        #
+        #    --write-out appends the status code on its own trailing line, so
+        #    the failure branch can name the status. The body itself is never
+        #    printed.
+        #
+        #    `|| true` so a transport failure reaches the failure branch and is
+        #    reported as HTTP 000, rather than aborting on curl's exit code
+        #    with no explanation.
+        SAUCE_RESP=$(curl -sS --write-out '\n%{http_code}' "$VENDOR_URL" \
```
charliesantos, line 104: Cross-checking this against the vendor in twilio-internal/video-sdk-e2e-credential-vendor#7: the request shapes are all correct (`get-saucelabs-credentials` takes no params and returns `username`/`accessKey`, `mint-voice-token` returns `token`, ttl 3600 is the ceiling, the identity is ~65 chars against a 128 cap). But three claim checks in that PR's `verifyOidc.private.js` each reject the tokens these jobs mint, given the values its `.env.example` and README prescribe for this repo's Service: `job_workflow_ref` must equal `.../workflows/e2e.yml` and every call here comes from `ci.yml`; `GITHUB_ALLOWED_REFS` is `refs/heads/main,main` while `ci.yml` triggers on every `pull_request` and this PR's base is `expo_support_1.x`; and `runner_environment` must be `github-hosted`, which is hardcoded, while the comment 70 lines up says the runner group is self-hosted.

Since CI is green, the deployed Service already diverges from what that PR documents - so fix its `.env.example` and README, or a redeploy from the template breaks this pipeline with a generic `401`. The third one needs a direct answer, because it isn't configurable: either `ubuntu-x64`/`macos-arm` are GitHub-hosted larger runners and the self-hosted comments here are wrong, or `DISABLE_OIDC_CHECK` is set on a deployed Service, which that README says never to do.

test/appium-harness/src/test-suites/voice-api.ts
```diff
@@ -163,17 +183,28 @@ const STEPS: Array<Step<Voice>> = [
...
+              .map((reportedDevice) => reportedDevice.uuid)
+              .sort(),
+            `the device uuids after selecting "${audioDevice.name}"`,
+          ).toStrictEqual(initialUuids);
```
charliesantos, line 199: This assertion and the comment at line 157 contradict each other. The comment says selecting regenerates every uuid on Android, `voice-api-test` is in the CI `SUITES` allowlist for both Android jobs, and a thrown expectation becomes the suite's failure status - so if the comment is accurate these jobs should be red.

"Does not fail in CI in practice" leaves it unresolved either way: either Android no longer regenerates uuids, in which case delete the KNOWN FAILING comment and the VBLOCKS-7133 reference rather than carrying them forward, or the assertion isn't executing on Android and the suite is green on a step that checks nothing. The Android step summary from the green run settles it.

test/appium-orchestrator/package.json
```diff
@@ -5,14 +5,12 @@
   },
   "dependencies": {
     "appium": "^3.3.1",
+    "appium-uiautomator2-driver": "^8.6.4",
+    "appium-xcuitest-driver": "^11.0.0",
```
charliesantos: This is a major-version downgrade of the driver that runs every iOS suite, and neither the description nor a comment mentions it. `setup.mjs` requests `'appium:platformVersion': '26'`, so: does v11 support iOS 26 on Sauce Labs, and was v12 tried first?

If the downgrade is what made the iOS jobs work, say so in the description and in a comment next to the pin, naming the v12 failure. Otherwise revert it - as written the next `yarn up` moves it back and nobody will know why the pipeline broke. (Moving both drivers from `devDependencies` to `dependencies` in the same change is right and worth keeping.)

charliesantos (changes requested):
Solid CI wiring and the credential handling is careful, but I can't sign off until the vendor-side OIDC question below is answered. Please link the green CI run in the description - two of my comments turn on it, and the `Test evidence` section currently has none.

On postcss (you asked): pin all three to one version. The whole value of the pin is that it reaches the code that gets built, and right now the root pin reaches only the install that produces no device build, so the description reads as fixed when the harness still resolves 8.5.26. One version, three places, or drop the root pin and say the CVE is unaddressed in the harness projects.

(8 findings from my local pass are omitted as already raised by Kevin or Copilot.)

.github/actions/saucelabs-upload/action.yml
kpchoy: True, your call on whether or not we add it.

.github/actions/e2e-credentials/action.yml
mhuynh5757: The deployed credential vendor has a testing `.env` published that just adds the `expo_support_1.x` branch in `GITHUB_ALLOWED_REFS`. It was only for testing. My plan was to just show that the CI worked with it, and then tear it down and republish under the VBlocks service account where the `GITHUB_ALLOWED_REFS` is correctly set to `refs/heads/main,main`. `DISABLE_OIDC_CHECK` was never set to `true` in any deployed serverless Vendor-Function.

test/appium-orchestrator/package.json
mhuynh5757: The regression back to v11 was a mistake. The drivers aren't actually used at all when connecting to Sauce Labs, so it wasn't a question. I will upgrade to v12.

.github/actions/saucelabs-upload/action.yml
mhuynh5757: Would prefer to hold off on this for now. Genuine hole, but I think very, very small and not a realistic issue in practice.

.github/actions/run-e2e-suites/action.yml
mhuynh5757: Would prefer to keep it safe and simple and skip logging `detail`.

